PR #23, opened 2015-11-14
Update schedule.py

shaqowski:
change cal_url to correct


Narcolapser:
Thank you very much for catching that change.

PR #38, opened 2016-10-20
Add event settings for timezones and type of event

Yenthe666:
Fix: Function description is SetAttendees while the function was named SetAttendee. Renamed the function to match the description
Add: Ability to set the timezone for the start of the meeting
Add: Ability to set the timezone for the end of the meeting
Add: Missing import for group, this is called by event.py to create a group object.
vent makes use of the group module on line 333 of this file. It was never imported and so that caused an error. Added group to the imports, this should relieve the issue.

### DIFF
--- a/O365/event.py
+++ b/O365/event.py
@@ -1,4 +1,5 @@
 from O365.contact import Contact
+from O365.group import Group
 import logging
 import json
 import requests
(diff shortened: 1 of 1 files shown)


Narcolapser:
thank you.

Issue #42, opened 2016-11-07
TabError: inconsistent use of tabs and spaces in indentation


fredottenhof:
I get this error after having installed O365 through pip:

>>> import O365
Traceback (most recent call last):
  File "<pyshell#0>", line 1, in <module>
    import O365
  File "C:\Users\fredo\AppData\Local\Programs\Python\Python35-32\lib\site-packages\O365\__init__.py", line 23, in <module>
    from .cal import Calendar
  File "C:\Users\fredo\AppData\Local\Programs\Python\Python35-32\lib\site-packages\O365\cal.py", line 7, in <module>
    from O365.event import Event
  File "C:\Users\fredo\AppData\Local\Programs\Python\Python35-32\lib\site-packages\O365\event.py", line 305
    def setStartTimeZone(self,val):
                                  ^
TabError: inconsistent use of tabs and spaces in indentation
>>> 

Please let me know when corrected

Fred Ottenhof

Narcolapser:
Yea. I saw that error. I forgot to check some one's pull request before I merged. I'll tend to it shortly.

Narcolapser:
Fixed it. update to 0.9.2 and you should be good.

fredottenhof:
Now that is speedy !!  Thanks very much !!

I’ll start the update immediately.

Fred.

PR #44, opened 2016-11-08
Add try/except for JSON body key

sampellino:
*NOT TESTED*

In regards to https://github.com/Narcolapser/python-o365/issues/43, I added the same try/catch that is in setBody, which will set the "Body" JSON key if it is not already set.


Narcolapser:
Very good. Before I merge, did you use tabs? I don't want to repeat #42.

sampellino:
I used GitHub's browser editor, and I used tabs with it, yes.

PR #47, opened 2016-12-07
Ability to pull more than 10 events

lbrictson:
The calendar getEvents function is only able to bring in 10 events.  Modifying the url to add &$top=50 will allow 50 (which is the max per page for the 0365 api)

### DIFF
--- a/O365/cal.py
+++ b/O365/cal.py
@@ -23,7 +23,7 @@ class Calendar( object ):
 		events_url - the url that is actually called to fetch events. takes an ID, start, and end date.
 		time_string - used for converting between struct_time and json's time format.
 	'''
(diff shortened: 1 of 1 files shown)


Narcolapser:
Yea. That's why I hadn't merged this yet. I was trying to think what would be the best route. I like the ability to catch more than 10 events, but I don't want it to be hard coded in. I certainly can't change the default or it might break other people's code. Is there a way you can get it done via an init argument or an a method call that changes the url? Cheers.

lbrictson:
Added it as a param and set the default value to 10 so it should no longer break code that was always looking for a max of 10 from previous versions, but you can now set it to a higher value.

Narcolapser:
Sorry for the delay. Busy December. Thank you sir!

PR #52, opened 2017-02-14
Added getID and getReceivedDate to message.py

ajwillo:
Added getID and getReceivedDate to message.py


Narcolapser:
Looks good. sorry for the delay. Thanks!

PR #55, opened 2017-07-05
Mainly added the option to add CC/Bcc to messages.

alexnbferreira:
I messed up on the commits though so it shows 100% of the file is changed. 
Great library by the way! Helped me a lot :)


Narcolapser:
I'm just glad to hear it's being used! It certainly has become my most popular library.

Thanks for adding cc and bcc, a number of people were wanting that feature.

Issue #16, opened 2015-05-29
Add OAuth2 authentication support


joincamp:
In order to allow a shared usage of this library (webapps), I suggest using [Microsoft's OAuth workflow](https://msdn.microsoft.com/en-us/office/office365/howto/common-app-authentication-tasks) for authentication.

Narcolapser:
While I agree with you I'm not entirely certain I can do this without having access to an Azure AD, which I do not. I'm not a webapp developer so OAuth is not something I'm very familiar with. Mind elaborating a little on what you imagine this would look like?

joincamp:
If you really want this feature I can help with it. I can't say that I particularly have a need for it, so you might want to follow YAGNI and wait until someone votes and says that they need this feature. I'm not actually a user, just trying to give some feedback. Do you still want some direction?

Narcolapser:
Ah, thanks for the feedback then. It is much appreciated. Like I mentioned in my reddit post this is my first project that has been worth note. So I am still trying to figure out how much features like these are things I /must/ do if I want this project to take off or not. I'm going to leave this for now and focus on some of the features that I was wanting, such as OneDrive files, and may come back to this later.

Narcolapser:
Two examples I found of OAuth in use:
* [python_clientcred](https://github.com/jasonjoh/python_clientcred)
* [pythoncontacts](https://github.com/jasonjoh/pythoncontacts)

Something I've considered doing is refactoring how Authentication works in O365 in general. Rather than having to pass auth around from class to class have, have an Auth class which would wrap requests and deal with authentication, be it OAuth2 or (Username,Password). Does that sound insane or like a reasonable idea to you?

roycem90:
Party Time, this is available in the latest code, give it a try

olaor:
This works for me. I guess this issue could be closed?

Narcolapser:
Thank you all.

PR #66, opened 2017-11-14
New Inbox and connection

roycem90:
1. Connection Class - Single point reference to any authentication, so no need to pass auth info everywhere and it allows configuring proxy as well

2. FluentInbox Class - A new format of inbox, with extra option, please have a look at readme.md for example code

3. fluent_message.py - Just added as a reference to enable the Connection based changes (right now its just copy of message.py with few additions, will be adding up stuff soon - if you like my way of coding)


Narcolapser:
Interesting. I'll have to consider this. I think I like the idea but I might want to apply it to the other objects and integrate it into the existing inbox instead of having separate classes. I'll try to get back to you sooner than later, but probably not before thanks giving.

roycem90:
i added it as another class, to make sure it doesn't effect existing users (as the api is completely different).. (^_^)

janscas:
This is great @roycem90.

A big problem here is that Basic Auth is deprected on November 2018. Any ideas on how to implement the new Oauth authentication?

Narcolapser:
Really? Basic Auth is going away? well, we will really have to get on issue #16. eep!

roycem90:
Thats bad, will try to look at oauth this week

Narcolapser:
how does fluent inbox handle multiple inboxes?

roycem90:
not sure i understand the question, what do u mean by multiple inboxes?
multiple instances of the inbox class? or multiple mail id inbox (using same creds)?

roycem90:
working on oauth2
code working as of now, will push up after some testing

janscas:
@roycem90 Anywhere where I can see the code with oauth2 working?

roycem90:
sorry, my laptop got a bunch of problems, just now got it fixed, i just commited the code on my master branch (u have the readme too)

janscas:
Ok, but it is not using the Connection class either. It is using the old auth

roycem90:
Shit looks like i overwrote the changes :(, but it was working when i tested.. 

Let me switch back my os and check

roycem90:
okie, my mistake
1. there was a minor bug, fixed it
2. the message is still not updated to use the whole connection thing, but it works as of now

Below is the code i used and it got the message subjects

from O365 import Connection, FluentInbox

client_id = <YOUR_CLIENT_ID>
client_secret = <YOUR_CLIENT_SECRET>

Connection.oauth2(client_id, client_secret)

# Create an inbox reference
inbox = FluentInbox()

# Fetch 20 messages from "Temp" folder containing "Test" in the subject
for message in inbox.from_folder('Inbox').search('Subject:Test').fetch(count=20):
    # Just print the message subject
    print(message.getSubject())

Narcolapser:
Would you say this is ready to merge in as is? I've looked it over and it seems fine. It looks like you've danced very carefully around the existing code to avoid breaking backwards compatibility. If so I think I'll go ahead an merge it and push a new version to pypi. I have some ideas I'd like to discuss with you on the topic of this. I'll send you a PM in a bit.

roycem90:
Yup, its ready to be merged, no impact on the existing code 💯

Narcolapser:
Github removed private messaging. oh well, I made a gist instead. Which will probably be better, this way if anyone wants to join the conversation they can. 

https://gist.github.com/Narcolapser/af9cb33e735049819c54b71de6b4963a

janscas:
I'll like to contribute and find the structure you detail on the gist very suitable.
@roycem90 in the connection.py, future.utils is only python2. This should work on python 3.

roycem90:
@janscas, ya i added that only for compatibility with python 2

Narcolapser:
Alright. that took a little more than expected to merge because the web editor was broken for some reason. *shrug* anyway. It's in now. Thanks for the contribution.

@both of you. I'll make a new milestone and open issues for working on this. I'm glad to have both of you contributing. I'll try to get on it as soon as possible. But you know, life never goes as planned. xD

roycem90:
👍

Issue #72, opened 2018-01-09
attachment.getType()


megadevx:
This call on a normal attachment is not getting the file type. I have been seeing the following:
ValueError: substring not found exception.
One might add a small try except that handles this.

megadevx:
Added a pull request with a proposed fix for this issue.

Narcolapser:
Merged pull request. I'll upload a new version to pip later today.

paul90046:
This the error message in ipython:

 400 Client Error: Bad Request for url: https://graph.microsoft.com/v1.0/me/messages?%24top=25&%24filter=attachments+eq+%272%27

I tried building it in graph explorer and I got:

https://graph.microsoft.com/v1.0/me/messages?%245top=25&%filter=eq%28attachments%2C+%272%27%29

this is for API 1.0

There seems to be a difference in how the url get assembled for contains versus equals.

janscas:
Can you post the filter not enconded in a url?

PR #73, opened 2018-01-09
Exception handling (ValueError) in getType()

megadevx:
Attempt to fix issue #72 with the Value Error. More testing needed.


Narcolapser:
Good work. Thanks.

Issue #75, opened 2018-01-10
attachment.save() not able to be run on windows.


megadevx:
the attachment.save() member function is unable to be run on windows because it uses a '/'.
This is a simple fix of using the os.path.join() to add the correct slash mark depending on the system the code is run on.

Narcolapser:
Good idea, can you put in a pull request?

megadevx:
Yep I just added the fix to a pull request #76.

PR #76, opened 2018-01-10
Fixed save() to work on windows and linux

megadevx:
imported path from os and used the path.join() function to get the correct path location for ones particular OS. This will make scripts written with this module also cross OS. This is a fix for issue #75.


Narcolapser:
I developed this almost exclusively on linux, some of these sorts of things are easy to miss. Thanks for the contribution!

PR #81, opened 2018-02-12
replace reciev -> receiv (spelling fix)

eseiver:
corrects all instances and variations of 'recieve' to 'receive'.
affects `Message` attribute "reciever" but is otherwise just in docstrings.


Narcolapser:
I would take the blame for these as a dyslexic, but it doesn't look like most of it is my code. ha! So i'm not the only one who struggles at times. xD Thanks for the corrections!

Issue #86, opened 2018-02-27
tab/spaces update has some errors


fcollman:
For example.. got this error on an automated build process that made use of this library.

error	27-Feb-2018 09:00:21	  File "/data/notify.py", line 1, in <module>
error	27-Feb-2018 09:00:21	    from O365 import Message
error	27-Feb-2018 09:00:21	  File "/opt/conda/lib/python2.7/site-packages/O365/__init__.py", line 26, in <module>
error	27-Feb-2018 09:00:21	    from .inbox import Inbox
error	27-Feb-2018 09:00:21	  File "/opt/conda/lib/python2.7/site-packages/O365/inbox.py", line 60
error	27-Feb-2018 09:00:21	    try:
error	27-Feb-2018 09:00:21	    ^
error	27-Feb-2018 09:00:21	IndentationError: unexpected indent

Narcolapser:
uploaded new version, thanks for the fix.

PR #87, opened 2018-02-27
fixing tabs in inbox.py

fcollman:
should address at least the specific problem reported in #86, larger scale checks should be done.


Narcolapser:
I should have looked closer. thanks.

PR #99, opened 2018-07-03
Added support for sending from shared mailbox

hellbe:
@Narcolapser another one :) 

* Send as/from shared mailbox
* Some readme notes on above and `FluentInbox` shared inbox support
* My editor trimmed some trailing whitespace automatically..

Again, would be awesome if we could get a new release on PyPI 🚢


Narcolapser:
Thanks, this API certainly has grown since I first started working with it in 2015. Nice to see other people catching these new features.

I'll see if I can get a new PyPI release out. I haven't tried it since they revamped the whole system and I'm about to go on vacation, so it may take a bit.

PR #103, opened 2018-07-11
Add the ability to specify default headers used in future requests.

Psynbiotik:
Add the ability to specify default headers used in future requests. For example this gives you the ability to set Connection.default_headers = {'Prefer': 'outlook.body-content-type="text"'}, to request to be sent text instead of html from outlook for future calls.

https://msdn.microsoft.com/en-us/office/office365/api/mail-rest-operations


Narcolapser:
Thank you.

PR #107, opened 2018-07-12
Have values of type dict returned also be converted to type MicroDict.

Psynbiotik:
This will ensure that results returned are always of never of type dict and instead are of type MicroDict.

I ran into an issue using getSenderEmail where 'EmailAddress' failed because the dict would not recognize the element 'emailAddress' as the same. By converting dict's to MicroDict's this method works for me again.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -17,6 +17,8 @@ class MicroDict(dict):
 		result = super(MicroDict, self).get(key[:1].lower() + key[1:], None)
 		if result is None:
 			result = super(MicroDict, self).get(key[:1].upper() + key[1:])
+		if type(result) is dict:
+			result = MicroDict(result)
 		return result
 
 


Narcolapser:
Sorry for the delay. I'll have to look at this one so I understand all the implications. I'm concerned about what happens when the payload for the API call is constructed. If the API requires "EmailAddress" and the dict provides "emailAddress" this could result in the library being broken. So I just want to make sure there are no such problems before I accept this. Can you verify that?

Psynbiotik:
Your concern is exactly what happened to me without this change when receiving results. Previously the json type is MicroDict but none of the sub objects are. So I added a line which will convert returned dicts to MicroDicts.

The only difference between a MicroDict and a dict is the __getitem__ call, a MicroDict attempts to resolve something like 'EmailAddress' as 'emailAddress' first, if that fails, then it tries 'EmailAddress', basically trying the first character as lower, then upper if that fails.  

The MicroDict doesn't actually modify any keys in the dictionary, just tries to get them using upper and lowercase variants, therefore it shouldn't have any affect on requests sent to APIs.

Narcolapser:
Gotcha, Thanks.

PR #108, opened 2018-07-12
Use oauth in message calls

Psynbiotik:
Save oauth in message for later use.
Messages now use oauth if it's been set.
Use response.ok to determine if call succeeded.
Changed Content-type to Content-Type.

Response doc: http://docs.python-requests.org/en/master/api/#requests.Response

### DIFF
--- a/O365/fluent_message.py
+++ b/O365/fluent_message.py
@@ -45,7 +45,7 @@ class Message(object):
 	draft_url = 'https://outlook.office365.com/api/v1.0/me/folders/{folder_id}/messages'
 	update_url = 'https://outlook.office365.com/api/v1.0/me/messages/{0}'
 
-	def __init__(self, json=None, auth=None, verify=True):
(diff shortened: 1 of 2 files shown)


O365/fluent_message.py
```diff
@@ -67,16 +67,16 @@ def __init__(self, json=None, auth=None, verify=True):
...
+		self.oauth = oauth
 
 	def fetchAttachments(self,**kwargs):
 		'''kicks off the process that downloads attachments locally.'''
 		if not self.hasAttachments:
 			log.debug('message has no attachments, skipping out early.')
 			return False
 
-		response = requests.get(self.att_url.format(
+		response = (self.oauth, requests)[self.oauth is None].get(self.att_url.format(
```
Narcolapser: Slick.

Narcolapser:
Thank you for your work.

PR #111, opened 2018-07-27
Message use 2.0 urls

Psynbiotik:
Use 2.0 version of the url if using OAuth, as the old urls don't work with an OAuth token using 2.0 scope


Narcolapser:
Thank you.

PR #123, opened 2018-09-17
Add __str__ method to cal

mbanders:
This way a user can simply print a calendar name.


Narcolapser:
Small but useful. Thanks.

PR #124, opened 2018-09-19
Improved subfolder navigation

jasonlopez01:
Added parent_id as argument to FluentInbox.from_folder method, added FluentInbox.folder attribute to store current folder and make it easier to get its folder info, like Id. Added example file illustrating how to navigate and get messages from inbox down to a nested subfolder.


Narcolapser:
You rock!

PR #136, opened 2018-11-06
Tests

Narcolapser:
Functional tests to ensure we don't have any regressions.


janscas:
Hi there, I'll create a project board with ideas, todo's, etc. Change it as you want.

Narcolapser:
Tests do work now so I am going to merge this so that we can more easily work on the tests together. It's certainly going to take some group effort to get 100% coverage.

PR #140, opened 2018-11-14
Docs

roycem90:
Only added some, but just want you @Narcolapser  get the setup done in settings, so we can see how it looks..

More coming..


Narcolapser:
Ok, I think I've flipped the appropriate switches. I've tentatively selected Slate as the documentation theme. If anyone has a strong opinion on the theme feel free to suggest it. I literally just looked at the ten and thought "eh, that looks nice." So I'm open to suggestions.

roycem90:
what theme?
this docs comes with html in readthedocs format

roycem90:
Settings > Github Pages > Source > master branch /docs
use the /docs not the master

after u change it have a look at https://o365.github.io/python-o365/

Narcolapser:
Apparently theme only matters if your documentation is nothing but a readme.

PR #141, opened 2018-11-15
Old api

Narcolapser:


Narcolapser:
Kept this for historical reasons.

PR #142, opened 2018-11-15
Auto deploy

Narcolapser:
Setting up an auto deployment process.


Narcolapser:
Looks like I just need to push this in. I'll try to get it to push a release after that.

PR #155, opened 2018-12-18
Sharepoint features

AnoojNair:
Adding get_listbyname to sharepoint


O365/sharepoint.py
```diff
@@ -185,6 +186,23 @@ def get_lists(self):
         data = response.json()
 
         return [self.list_constructor(parent=self, **{self._cloud_data_key: lst}) for lst in data.get('value', [])]
+    
+    def get_listbyname(self,displayName):
```
Narcolapser:
```suggestion
    def get_list_by_name(self,displayName):
```
We are using [pep8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names) as our style guide. That requires every word in a function name to be separated by an underscore.

Narcolapser:
@janscas This looks good to me. But you know the sharepoint code better. Just wanted you to double check it before we merge.

O365/sharepoint.py
```diff
@@ -99,7 +99,8 @@ class Site(ApiComponent):
 
     _endpoints = {
         'get_subsites': '/sites',
-        'get_lists': '/lists'
+        'get_lists': '/lists',
+        'get_listbyname':'/lists/displayName'
```
janscas: As we are following PEP8, displayName should be display_name, and get_listbyname should be get_list_by_name.

Also, we are using format for string replacement. so instead of '/lists/display_name' use '/lists/{display_name}' so later you can format this path instead of using 'replace'

O365/sharepoint.py
```diff
@@ -185,6 +186,23 @@ def get_lists(self):
         data = response.json()
 
         return [self.list_constructor(parent=self, **{self._cloud_data_key: lst}) for lst in data.get('value', [])]
+    
+    def get_list_by_name(self,displayName):
```
janscas: display_name

O365/sharepoint.py
```diff
@@ -185,6 +186,23 @@ def get_lists(self):
...
+        """ Returns a sharepoint list based on the display name of the list 
+        """
+        
+        if not displayName:
+            raise ValueError('Must provide a valid list display name')
+
+        url = self.build_url(self._endpoints.get('get_listbyname').replace('displayName',displayName))
```
janscas: url = self.build_url(self._endpoints.get('get_list_by_name').format(display_name=display_name))

O365/sharepoint.py
AnoojNair: Thankx , i have made the change 👍

Narcolapser:
@AnoojNair There appears to be a merge conflict now. Will github allow you to resolve that? If not I'll try to do it. I'd just prefer some one who knows what's new and old to do it.

O365/sharepoint.py
```diff
@@ -93,6 +115,20 @@ def get_items(self):
         return [self.list_item_constructor(parent=self, **{self._cloud_data_key: item})
                 for item in data.get('value', [])]
 
+    def get_item_by_id(self,id):
```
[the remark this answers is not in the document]
AnoojNair: @janscas i didn't know that.. Thanks .  i have updated it .

O365/sharepoint.py
```diff
@@ -25,14 +25,36 @@ def __init__(self, *, parent=None, con=None, **kwargs):
 
         cloud_data = kwargs.get(self._cloud_data_key, {})
 
-        self.object_id = cloud_data.get('id')
+        self.id = cloud_data.get('id')
```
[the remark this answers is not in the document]
AnoojNair: Done. Thankx

AnoojNair:
@Narcolapser , Looks like i can't resolve the conflicts(seems to be an access issue) .

janscas:
@AnoojNair What's the return value of the request inside the function `get_list_by_name`?

It's a list or a single SharepointListItem?

AnoojNair:
> @AnoojNair What's the return value of the request inside the function `get_list_by_name`?
> 
> It's a list or a single SharepointListItem?

@janscas  it's a list

PR #159, opened 2018-12-22
Planner and tasks

AnoojNair:
PR for planner related features like tasks

### DIFF
--- a/O365/__init__.py
+++ b/O365/__init__.py
@@ -13,5 +13,6 @@ from .drive import Storage, Drive, Folder, File, Image, Photo
 from .mailbox import MailBox
 from .message import Message, MessageAttachment, Recipient
 from .sharepoint import Sharepoint, Site
+from .planner import Planner,Task
 from .utils import ImportanceLevel, Query
 from .utils import OneDriveWellKnowFolderNames, OutlookWellKnowFolderNames
--- a/O365/account.py
+++ b/O365/account.py
@@ -6,6 +6,7 @@ from O365.drive import Storage
 from O365.mailbox import MailBox
 from O365.message import Message
 from O365.sharepoint import Sharepoint
+from O365.planner import Planner
 from O365.utils import ME_RESOURCE
 
 
@@ -153,3 +154,13 @@ class Account(object):
                 'Sharepoint api only works on Microsoft Graph API')
 
         return Sharepoint(parent=self, main_resource=resource)
+
+    def planner(self, *, resource=''):
+        """ Get an instance to read information from Microsoft planner """
+        
+        if not isinstance(self.protocol , MSGraphProtocol):
+             # TODO: Custom protocol accessing OneDrive/Sharepoint Api fails here
+            raise RuntimeError(
+                'planner api only works on Microsoft Graph API')
+
+        return Planner(parent=self, main_resource=resource)
--- a/O365/planner.py
+++ b/O365/planner.py
@@ -1,0 +1,110 @@
+import logging
+
+from dateutil.parser import parse
+
+from O365.address_book import Contact
+from O365.drive import Storage
+from O365.utils import ApiComponent
+
+log = logging.getLogger(__name__)
+
+class Task(ApiComponent):
+    """ A Microsoft Planner task """
+
+    _endpoints = {
+    }
+
+    def __init__(self, *, parent=None, con=None, **kwargs):
+        """ A Microsoft planner task
+
+        :param parent: parent object
+        :type parent: Sharepoint
+        :param Connection con: connection to use if no parent specified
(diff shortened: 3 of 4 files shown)


roycem90:
@AnoojNair i hope this has been tested?

Narcolapser:
It is the right scope so we'd be happy to include the code. I downloaded it and fiddled with the code some myself. Unfortunately I am not an administrator so I cannot consent to the permissions needed to use this portion of the API. So I can't speak to it's functionality, but it looks right.

Because it is not intuitive can you include in the comments some where what [permissions](https://docs.microsoft.com/en-us/graph/api/planneruser-list-tasks?view=graph-rest-1.0) are needed to get this portion of the API to work?

Bonus points if you can write some tests for this class. Cheers!

janscas:
As @Narcolapser said, I also can't test this because I can't set app permissions (just delegated permissions).

But the code looks fine as @Narcolapser  said.

AnoojNair:
@Narcolapser @janscas @roycem90  . Thanks for the feedback. I will include tests and documentation into the PR as well

AnoojNair:
@Narcolapser , Just wanted to check how the o365_token.txt gets populated? i wanted to check what would be the minimum permission required to read user's tasks. i feel it should be 'Tasks.Read' . 

Each time i authenticate, the token.txt file has all these scopes

"https://graph.microsoft.com/Group.Read.All",
  "https://graph.microsoft.com/Mail.ReadWrite",
  "https://graph.microsoft.com/Mail.Send",
  "https://graph.microsoft.com/Sites.Read.All",
  "https://graph.microsoft.com/Tasks.Read",
  "https://graph.microsoft.com/User.Read"

i tried removing them all and just keeping Tasks.Read. and authenticate again using 
result = acc.authenticate(scopes=['Tasks.Read'])

then it re-populates back to

 "scope": [
  "profile",
  "openid",
  "email",
  "https://graph.microsoft.com/Group.Read.All",
  "https://graph.microsoft.com/Mail.ReadWrite",
  "https://graph.microsoft.com/Mail.Send",
  "https://graph.microsoft.com/Sites.Read.All",
  "https://graph.microsoft.com/Tasks.Read",
  "https://graph.microsoft.com/User.Read"
 ]

Please let me know if i am missing something. i haven't adding any permission in the registration portal as of yet.

image

janscas:
I think that’s because you have previously consent that app with those scopes. 

Remove the app and make it again. You will have to change the client id and client secret.

AnoojNair:
Thanks @janscas . I am trying to write some unit tests for planner, but when i run py.test i get the modulenotfounderror for O365. Am i missing something?

image

roycem90:
try installing it 
(run this from your git repo, that will install O365 by linking to your code location)
pip install -e ./

AnoojNair:
Thanks @roycem90 . Also guys @janscas @Narcolapser . I have included tests for planner .But the return value is always None. I am not sure how to complete the oauth flow while the test executes as the oath flow requires me to paste the url from the browser.

Please let me know if am missing something in the test class.

roycem90:
@AnoojNair u need not get the data from original account (that part is separate you do a manual test for those stuff, that is what we do most of the time when we make changes)

these tests that you are adding are automated py tests.. we generally create mocks 
from unittest.mock import MagicMock and mock patching or 
manual mocking, below is how we do manual mock

- Get a sample response (probably the one here - https://docs.microsoft.com/en-us/graph/api/planneruser-list-tasks?view=graph-rest-1.0)
- Save it to a planner_my_tasks.json under tests/samples dir

`
    import json

    class MockResponse:
        resp_file = None

        def json(self):
            with open(MockResponse.resp_file) as f:
                resp_json = json.load(f)
            return resp_json

    class MockConnection:
        def get(self, url, params=None, **kwargs):
            self.url = url
            self.kwargs = kwargs
            return MockResponse()
    
   
    def test_get_my_tasks(self):
        MockResponse.resp_file = 'planner_my_tasks.json'
        # after this as your connection returns a valid json, your planner class should do the rest
`

AnoojNair:
Thanks @roycem90 , Are there any sample mock tests in the project that i can check and create mock test similarly?

roycem90:
don't think we wrote any tests properly, you can use that code i added above i guess it works, or just to do manual tests and push the code once you are sure all is good :)

roycem90:
can you do a commit commenting out the test (that is failing, would like green :) )

@janscas @Narcolapser code seems fine, i m ok to merge this

janscas:
Seems ok to me also

nguyenl95:
can anyone write unittests for this?

janscas:
Instead of asking, do it yourself

PR #162, opened 2019-01-06
Updated calendar example code 

nickyspag:
The updated calendar example in the README was not working for me (400 bad request) - I changed to first call get_default_calendar and then call new_event() on that and that worked - also kicked it into the future by adding a year :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,17 +498,18 @@ import datetime as dt
 # ...
 schedule = account.schedule()
 
-new_event = schedule.new_event()  # creates a new event in the user default calendar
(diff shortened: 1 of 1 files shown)


janscas:
I have tried and it works for me. In fact calling new_event on the schedule object is the same as calling it from the default calendar.

This works:

```python
# ...
schedule = account.schedule()

event = schedule.new_event('test!')
event.start = dt.datetime(2019, 1, 10)
event.save()
```

@nickyspag  what action is giving you a bad request?

PR #164, opened 2019-01-08
Make project's import paths relative

cfinnberg:


cfinnberg:
Hi @derrix060, in this case I changed it because I tried to use the project from a subfolder and absolute paths in imports doesn't allow that. Also I thought that relative paths was the norm now.

I now realize the change may has its downsides. I suppose it's up to project's maintainers to decide which is better to the project.

roycem90:
@janscas whats your take here? relative or absolute?

janscas:
Hi, that’s what PEP8 says about this:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
> 
> import mypkg.sibling
> from mypkg import sibling
> from mypkg.sibling import example
>
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
> 
> from . import sibling
> from .sibling import example

I prefer absolute imports just because explicit is better than implicit, but I really don’t mind in this case.

I really don’t understand why is @cfinnberg using this as a subfolder. It’s meant to be used as a library.

So, I vote to leave it as realtive imports for now and see in the future if we change it... but you have the last word @roycem90.

roycem90:
fine leaving it relative for now, we can go to absolute if we have any troubles in future

janscas:
@cfinnberg for this you should use `pip install -e /path/to/your/package`

With this the changes made to the clonned O365 package will become available in your main project.

cfinnberg:
@janscas thanks! I will try it

PR #176, opened 2019-01-18
Async connection using aiohttp

kellerza:
Added an async connection, using aiohttp

There will be quite a lot of shuffling around of code to get to async, since any sync call must be removed from the flow. (request.get / file load/save etc)

The Connection class was split into two parts - a transport agnostic, ConnectionBase, with the sync version being called Connection. In aio/connection.py, ConnectionBases is extnded with async capble methods. This approach seems good for Connection, since you dont want to force aiohttp to be installed if not required. The Apicomponents can follow another approach have methods with two different names, all async functions prepended with aio_).

This approach allows some re-use of ConnectionBase and does not force async on the project if you only import the normal connection.

Some examples (both webservers and clients, needs an Application with redirect to http://localhost:5000):
- examples/aio/sample_aiohttp.py 
- examples/aio/sample_aiohttp_OAuth2Session.py (uses the raw aio_http OAuth2Session class,- based off requests OAuth2Session

I introduced a single API component - Info - that has a sync and async interface (prepended with aio_). This approach however uses the async def / await syntax of Python3.5+, making that a minimum requirement.

This is a big change, to potentially create a smaller PR, we can move the `ConnectionBase` and `Connection(ConnectionBase)` classes to aio/ and keep the non-async version as is... until you are happy to for it to be moved into the main code. Open to suggestions.


kellerza:
Similar to [these](https://github.com/O365/python-o365/blob/master/tests/test_connection.py#L15)?

I don't mind adding unit tests with at least some mocks to the async part, but improving unit tests for existing functionality is a bit out of the scope of this PR

kellerza:
Thanks @janscas, I don't want to go too deep down the rabbit hole without knowing where we'd like to end up...

The main reason would not be to do 1000s of concurrent calls, but have `await` support in async applications (like webservers).

kellerza:
Started using the MSAL Python library for authentication.

See https://gist.github.com/kellerza/8aad3952086b827a9f32516373df1623 for an example

PR #179, opened 2019-01-23
Authentication & other improvements

janscas:
Improvements on authentication and other improvements

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import time
+import warnings
 from pathlib import Path
 
 from oauthlib.oauth2 import TokenExpiredError
@@ -16,7 +17,7 @@ from requests_oauthlib import OAuth2Session
 from stringcase import pascalcase, camelcase, snakecase
 from tzlocal import get_localzone
 
-from .utils import ME_RESOURCE
+from .utils import ME_RESOURCE, BaseTokenBackend, FileSystemTokenBackend, Token
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


janscas:
Creating this PR to group all the changes around the library authentication

janscas:
_(Pasting this here also)_

What I did is the following:

**- Oauth stuff:**

- [x] Created `account.is_authenticated` as a property. The "request" argument is no needed as we can get the expiration time on refresh tokens as well.
- [x] Modified `account.authenticate` (and all the other oauth flows) to allow users to pass scope helpers or raw scopes. Now you can do: `account.autheticate(scopes=['basic', 'message_all', 'https://graph.microsoft.com/Calendars.ReadWrite'])` and it will do the appropriate thing.
The only caveat is that if you create account with the protocol MSGRAPH and you pass Office365 scopes the api will not catch the error and the authentication will fail. But that's fairly easy to avoid.
- [x] The `connection.refresh_token()` now checks if the access token can indeed be refreshed and therefore will not try to refresh it. This method can now be called manually as well.

**- Token backends:**

- [x] Token backends are implemented. Now anyone can define a new token backend to store the token anywhere. For now only `FileSystemTokenBackend` is available.
- [x] Tokens are no longer normal python dicts. They are a subclass of dict called `Token`. It implements the following properties: "is_long_lived" to know if there is refresh token. "expiration_datetime" to retrieve the expiration datetime of the whole token (access and refresh tokens) and "is_expired" that will tell the library if the token has expired completely and therefore the user will have to authenticate again.

- [x] Updated Readme (I've improved the readme to adapt to these changes and also make it more understandable)

- [ ] Updated Docs: Pending

O365/connection.py
```diff
@@ -454,49 +458,71 @@ def get_session(self, token_path=None):
...
+
+        token = self.token_backend.token
+        if token and token.is_long_lived:
+            client_id, client_secret = self.auth
+            token = Token(self.session.refresh_token(
+                self._oauth2_token_url,
+                client_id=client_id,
+                client_secret=client_secret))
+        else:
+            log.error('You can not refresh an access token that has no "refreh_token" available.'
```
roycem90: really like the error message for offline_access, helps avoid few tickets :)

PR #185, opened 2019-02-04
add tenant_id parameter to Connection class

### LINKED ISSUE #184: Specify tenant-specific (directory) endpoint
I'm pretty new to this O365, Microsoft Graph, OAuth2 stuff. I registered an application and created a secret. Then I ran the following: 

```
from O365 import Connection, MSGraphProtocol, MSOffice365Protocol
from O365.account import Account
credentials = ("<client_id>", "<secret>")
scopes = ['https://graph.microsoft.com/User.Read']
protocol_graph = MSGraphProtocol()
scopes_graph  = protocol_graph.get_scopes_for(scopes)
con = Connection(credentials, scopes=scopes_graph)
con.get_authorization_url()
```

The result is an URL that points to `https://login.microsoftonline.com/common/oauth2/v2.0/authorize?response_type=code&client_id=<client_id>...redirect=...scope=...access_type=offline`. I opened that URL in a browser, successfully authenticated an account and got the following error from Microsoft:

> AADSTS50194: Application '<client_id>' is not configured as a multi-tenant application. Usage of the /common endpoint is not supported for such applications created after '10/15/2018'. Use a tenant-specific endpoint or configure the application to be multi-tenant.

I do not want to switch to multi-tenant. So I manually changed the URL and replaced `common` with the <directory_id> (also labeled `Directory (tenant) ID`). Then I no longer get the AADSTS50194 error.

Is there a parameter in `python-o365` to set the `direcory_id` value? If not and it makes sense then I could look into sending a PR.

frennkie:
Applications that are not set up for multi tenant must talk to a
different endpoint than "common". This adds an argument for this to
the Connection class (__init__).
Fix #184.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -252,17 +252,13 @@ class MSOffice365Protocol(Protocol):
 class Connection:
     """ Handles all communication (requests) between the app and the server """
 
-    _oauth2_authorize_url = 'https://login.microsoftonline.com/common/' \
-                            'oauth2/v2.0/authorize'
-    _oauth2_token_url = 'https://login.microsoftonline.com/common/' \
-                        'oauth2/v2.0/token'
     _allowed_methods = ['get', 'post', 'put', 'patch', 'delete']
 
     def __init__(self, credentials, *, scopes=None,
                  proxy_server=None, proxy_port=8080, proxy_username=None,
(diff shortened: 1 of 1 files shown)


janscas:
Thanks! very useful

PR #195, opened 2019-02-22
Added a storage sample 

rorytarabay:
I created this sample to help me understand how to use your module. Thought I'd pull in case it helps others.

### DIFF
--- a/examples/storageDownloadFile
+++ b/examples/storageDownloadFile
@@ -14,8 +14,8 @@ import pandas as pd
 from O365 import Account
 
 # Generated on the app registration portal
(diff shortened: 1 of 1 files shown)


janscas:
@rorytarabay Remove the app registered with that secret you publish here just to be secure.

rorytarabay:
> @rorytarabay Remove the app registered with that secret you publish here just to be secure.

Thanks! Such a silly mistake. The secret was changed within a minute, the app has been deleted, and the audit logs have been verified to ensure nothing nasty took place.

Thankfully all clear!

PR #196, opened 2019-02-22
Correcting CalendarColors names

FrostmourneHungers:
The "CalendarColor" names seem to start with a small not a capital "L".
To keep a consistent namescheme "MaxColor" and "Auto" should be renamed to "maxColor" and "auto", too.
( In row 1499 it is already "auto")

I encountered this problem when my code throwed the following error while trying to get a calendar:

```
>>> schedule = account.schedule()
>>> calendar = schedule.get_calendar(calendar_name='Heizung')

ValueError: 'lightGreen' is not a valid CalendarColors


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<pyshell#9>", line 1, in <module>
    calendar = schedule.get_calendar(calendar_name='Heizung')
  File "E:\Programme\Python\Python3\lib\site-packages\O365\calendar.py", line 1849, in get_calendar
    **{self._cloud_data_key: data})
  File "E:\Programme\Python\Python3\lib\site-packages\O365\calendar.py", line 1501, in __init__
    self.color = CalendarColors(color)
  File "E:\Programme\Python\Python3\lib\enum.py", line 309, in __call__
    return cls.__new__(cls, value)
  File "E:\Programme\Python\Python3\lib\enum.py", line 561, in __new__
    raise exc
  File "E:\Programme\Python\Python3\lib\enum.py", line 545, in __new__
    result = cls._missing_(value)
  File "E:\Programme\Python\Python3\lib\enum.py", line 574, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 'lightGreen' is not a valid CalendarColors
```

I'm using a private outlook account (wich is accessed through "https://outlook.live.com/") by the way.
Buissnes / school accounts are accessed through "https://outlook.office365.com/owa/" (i can't log in there with my private account).
Maybe Microsoft has an inconsistent color name convention here (difference between private and buissnes).
This should be checked before merging my pull request.

This is my first Pull request by the way.
I hope Iam doing this right 🤓

### DIFF
--- a/O365/calendar.py
+++ b/O365/calendar.py
@@ -50,17 +50,18 @@ class EventShowAs(Enum):
 
 
 class CalendarColors(Enum):
-    LightBlue = 0
-    LightGreen = 1
-    LightOrange = 2
-    LightGray = 3
-    LightYellow = 4
-    LightTeal = 5
-    LightPink = 6
-    LightBrown = 7
-    LightRed = 8
-    MaxColor = 9
-    Auto = -1
+    LightBlue = 'lightBlue'
+    LightGreen = 'lightGreen'
+    LightOrange = 'lightOrange'
+    LightGray = 'lightGray'
+    LightYellow = 'lightYellow'
+    LightTeal = 'lightTeal'
+    LightPink = 'lightPink'
+    LightBrown = 'lightBrown'
+    LightRed = 'lightRed'
+    MaxColor = 'maxColor'
+    Auto = 'auto'
+
 
 
 class EventType(Enum):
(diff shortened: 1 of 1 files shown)


janscas:
The Enum members name has nothing to do with that.
It's the value want counts.
It seems that the 'color' must be a string.

Can you update your pull request by leaving the Enum members as before but updating the members values to be strings like: 'lightGreen', etc...

Thanks!! Good PR

janscas:
Btw it seems the color names use capital camel case ('LightGreen' instead of 'lightGreen'), can you check what casing works?

Thanks

FrostmourneHungers:
While the documentation of Mircosoft Graph REST API v1 says its 

> 
calendarColor | Specifies the color theme to  distinguish the calendar from other calendars in a UI. The property  values are: LightBlue=0, LightGreen=1, LightOrange=2, LightGray=3,  LightYellow=4, LightTeal=5, LightPink=6, LightBrown=7, LightRed=8,  MaxColor=9, Auto=-1
-- | --

it acually returns 'lightBlue', 'lightGreen', ...

I also noticed that in line 1500 of calendar.py there was is TODO for the colors:
```
        color = cloud_data.get(self._cc('color'), -1)
        if isinstance(color, str):
            color = -1 if color == 'auto' else color			
            # TODO: other string colors?
        self.color = CalendarColors(color)
```

So i updated the file

janscas:
I don't like this approach. Just correct the Enum values and this will work fine without all this if-else blocks:

```python
class CalendarColors(Enum):
    LightBlue = 'lightBlue'
    LightGreen = 'lightGreen'
    # etc...
```

And then, when capturing the color:

```python
color = cloud_data.get(self._cc('color'), 'auto')  # defaults to auto instead of -1
self.color = CalendarColors(color)
```

This should work. To protect against undefined colors (not sure if this is posible) maybe add a try except block on self.color = CalendarColors(color) to set CalendarColors('auto') if the color is not found.

janscas:
That's great!

FrostmourneHungers:
That's definitely better :)
Thank you for guide me through my first PR.

PR #198, opened 2019-02-23
fix minor typos and PEP8 issues

frennkie:
Typo in `Function` - also code style fixes (which PyCharm complained about)


janscas:
The examples are not updated to the library version >= 1.0.

We must rewrite them all.

Issue #197, opened 2019-02-23
EventMessage - handle meeting invites


frennkie:
I am currently working on migrating a "Bot" that handles several tasks from our on-Prem Exchange (using the Exchange Web Service EWS) to Exchange Online/O365. For EWS I had been using https://github.com/ecederstrand/exchangelib - which is great.. but it appears that Microsoft is phasing out EWS.

I have been able to successfully authenticate using `python-o365` and I can get messages from the mailbox and also calendar entries from the calendar. What I am missing is a dedicated `EventMessage`. Meeting invitations (or rejections) that are in the inbox have the same type as emails in python-o365.

I am currently working with the `MSGraphProtocol` and also there `EventMessage` is a separate Type: https://docs.microsoft.com/en-us/previous-versions/office/office-365-api/api/version-2.0/complex-types-for-mail-contacts-calendar#eventmessage.

Am I'm doing something wrong or is this simply not implemented (yet?!)?

frennkie:
I looked into the raw API response and the following two fields need to be evaluated:

* `@odata.type` (e.g. `#microsoft.graph.eventMessage`)
* `meetingMessageType` (e.g. `meetingRequest`)

How can I access this in `python-o365`?

janscas:
We have this as a 'ToDo':

https://github.com/O365/python-o365/projects/2#card-14588162

This should be easy to implement:

- add the message type property
- add the meetingMessageType property
- add a method to retrieve the related event.

frennkie:
@janscas The `ToDo` is pretty brief.. Do you also consider [accepting](https://docs.microsoft.com/en-us/graph/api/event-accept?view=graph-rest-1.0)/[declining](https://docs.microsoft.com/en-us/graph/api/event-decline?view=graph-rest-1.0) a meeting request in scope? Those are the things I am currently using through EWS.

janscas:
That’s already implemented. Check the calendar.py code.

frennkie:
@janscas Great! :-D I didn't see that. Would it help if I start/send a PR?

frennkie:
I was able to retrieve the related event of a Event Message like so: 
`https://graph.microsoft.com/v1.0/me/messages/A...A=?$expand=microsoft.graph.eventMessage/event`

I noticed that some query key words are implement in utils.py - but not `expand`. So I assume it should be added there, right?

frennkie:
[...]

PR #199, opened 2019-03-02
add the expand query parameter

frennkie:
As discussed in #197 it make sense to have the query parameter `$expand` (mostly adapted from the $select parameter).

The `convert_case`  breaks the value of e.g. `microsoft.graph.eventMessage/event` (it results in `microsofrapventMessagvent`. I only tested this on `MSGraphProtocol`, so this will most likely need to be checked fixed for `MSOffice365Protocol`

### DIFF
--- a/O365/utils/utils.py
+++ b/O365/utils/utils.py
@@ -321,14 +321,36 @@ class Query:
         self._filters = []  # store all the filters
         self._order_by = OrderedDict()
         self._selects = set()
+        self._expands = set()
 
     def __str__(self):
-        return 'Filter: {}\nOrder: {}\nSelect: {}'.format(self.get_filters(),
-                                                          self.get_order(),
-                                                          self.get_selects())
+        return 'Filter: {}\nOrder: {}\nSelect: {}\nExpand: {}'.format(self.get_filters(),
+                                                                      self.get_order(),
+                                                                      self.get_selects(),
+                                                                      self.get_expands())
 
     def __repr__(self):
         return self.__str__()
+
+    @fluent
+    def expand(self, *relationships):
+        """ Adds the relationships (e.g. "microsoft.graph.eventMessage/event" or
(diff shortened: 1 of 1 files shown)


frennkie:
@janscas Feel free to edit or even discard if you have already done this separately/differently.. :-D

O365/utils/utils.py
```diff
@@ -321,15 +321,37 @@ def __init__(self, attribute=None, *, protocol):
...
+    @fluent
+    def expand(self, *relationships):
+        """ Adds the relationships (e.g. "microsoft.graph.eventMessage/event" or
+        "attachments") that should be expanded to the $expand parameter
+
+        :param str relationships: the relationships tuple to expand.
+        :rtype: Query
+        """
+
+        for relationship in relationships:
```
janscas: To control whether to add 'microsoft.graph.eventMessage' if the protocol is MSGrapProtocol or the 'Microsoft.OutlookServices.EventMessage' if you are using Outlook rest api, you have to call self.protocol.get_service_keyword. I'll implement this

janscas:
@frennkie This PR is a Draft so I can not merge it nor use the code. 
I'll commit the changes that include some of your code and some modifications

frennkie:
Feel free.

I was not aware that you can not edit this Draft PR.. First time I'm using it. :-D

frennkie:
@janscas I did a work-in-progress implementation of auto parse too.. Might be a reference for you: https://github.com/frennkie/python-o365/tree/auto-parse-event?files=1

janscas:
> did a work-in-progress implementation of auto parse too

What do you mean by 'auto parse'?

frennkie:
Have a look at it... If $expand was given then I parse the Event and add it the event attribute of Message.

janscas:
> Have a look at it... If $expand was given then I parse the Event and add it the event attribute of Message.

ok, this is already done in the commit

PR #201, opened 2019-03-02
fix broken URL for accepting an event

### LINKED ISSUE #200: Can't accept a Meeting (event)
I tried to accept a meeting request and it failed.
End of the error message:

```
  File "C:\Users\User\venv\lib\site-packages\requests_oauthlib\oauth2_session.py", line 384, in request
    raise InsecureTransportError()
oauthlib.oauth2.rfc6749.errors.InsecureTransportError: (insecure_transport) OAuth 2 MUST utilize https
```

Accepting `tentatively` works fine. But accepting normally fails. The cause is that the URL is lost in the `accept` code path. Instead of ` https://graph.microsoft.com/v1.0/me/events/A...A=/tentativelyAccept` only `/accept` is called.

It's an easy and small fix.. I will send a PR.

frennkie:
fix #200

### DIFF
--- a/O365/calendar.py
+++ b/O365/calendar.py
@@ -1388,7 +1388,7 @@ class Event(ApiComponent, AttachableMixin, HandleRecipientsMixin):
 
         url = self.build_url(
             self._endpoints.get('event').format(id=self.object_id))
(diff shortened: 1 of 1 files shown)


janscas:
Thanks!

PR #210, opened 2019-03-20
Ability to create, and edit listitems in sharepoint

### LINKED ISSUE #181: Possible to edit sharepoint lists?
Hi

Is it possible in this version to open and edit a sharepoint list? Is there an example for this?

Kind regards
Jens

jtweeder:
Took a stab on adding some functionality that I have working on my O365 sharepoint.  Open to feedback on any way to make this better, but figured it was a start.  Review at your leisure.

Added a crosswalk between display name, and internal name for user defined columns in the list.  Added update_list_item to update a given listitem, and create_list_item to create a new list item.

Additions:
1. SharepointList.column_name_cw - dictionary of {display_name: internal_name} for user defined columns in a sharepoint list. (I have found this useful when doing things like list updates, additions, and pulling down detail)

2. SharepointList.update_list_item(item_id, update) - given an item id and dict of {internal column name: updated data} it will update the sharepoint list item, and return the updated item.

3. SharepointList.create_list_item(new_data) - given a dict of {internal column name: new data} it will make a new item in the sharepoint list and return the new item as a SharepointListItem

Open to adding information to documentation as well, but wanted to get the code ironed out first.
Address and Closes #181

### DIFF
--- a/O365/sharepoint.py
+++ b/O365/sharepoint.py
@@ -189,7 +189,11 @@ class SharepointList(ApiComponent):
             self._cc('contentTypesEnabled'), False)
         self.hidden = lst_info.get(self._cc('hidden'), False)
         self.template = lst_info.get(self._cc('template'), False)
-
+        
+        #Crosswalk between display name of user defined columns to internal name
+        self.column_name_cw = {col.display_name: col.internal_name for 
+                               col in self.get_list_columns() if not col.read_only}
+        
     def get_items(self):
         """ Returns a collection of Sharepoint Items
 
@@ -236,8 +240,41 @@ class SharepointList(ApiComponent):
 
         return [self.list_column_constructor(parent=self, **{self._cloud_data_key: column})
                 for column in data.get('value', [])]
+    
+    def update_list_item(self, item_id, update):
+        """Updates a list item with a given update
+           
+        :param item_id: List item id to update
+        :param update: Data to update list item with {'col_name': 'Updated Data'}
+        
+        :rtype: SharepointListItem
+        """
+        
+        url = self.build_url(self._endpoints.get('get_item_by_id').format(item_id=item_id) + '/fields')
+        
+        response = self.con.patch(url, update)
+        
+        data = response.json()
+        
+        return self.get_item_by_id(item_id)
 
-
+    def create_list_item(self, new_data):
+        """Create new list item
+        
+        :param new_data: dictionary of {'col_name': col_value}
+                    
+        :rtype: SharepointListItem
+        """
+        
+        url = self.build_url(self._endpoints.get('get_items'))
+        
+        response = self.con.post(url, {'fields': new_data})
+        
+        data = response.json()
+        
+        return self.list_item_constructor(parent=self, **{self._cloud_data_key: data})
+    
+    
 class Site(ApiComponent):
     """ A Sharepoint Site """
 
--- a/O365/sharepoint.py
+++ b/O365/sharepoint.py
@@ -65,7 +65,7 @@ class SharepointListColumn(ApiComponent):
 
 
 class SharepointListItem(ApiComponent):
-    _endpoints = {}
+    _endpoints = {'update_list_item': '/items/{item_id}/fields'}
 
     def __init__(self, *, parent=None, con=None, **kwargs):
         """ A Sharepoint ListItem within a SharepointList
@@ -81,6 +81,7 @@ class SharepointListItem(ApiComponent):
         if parent and con:
             raise ValueError('Need a parent or a connection but not both')
         self.con = parent.con if parent else con
+        self._parent = parent
 
         # Choose the main_resource passed in kwargs over parent main_resource
         main_resource = (kwargs.pop('main_resource', None) or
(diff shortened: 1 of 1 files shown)


O365/sharepoint.py
```diff
@@ -236,8 +240,41 @@ def get_list_columns(self):
...
+    def update_list_item(self, item_id, update):
+        """Updates a list item with a given update
+           
+        :param item_id: List item id to update
+        :param update: Data to update list item with {'col_name': 'Updated Data'}
+        
+        :rtype: SharepointListItem
+        """
+        
+        url = self.build_url(self._endpoints.get('get_item_by_id').format(item_id=item_id) + '/fields')
```
janscas: This url should be added to endpoints.
The urls used must appear as endpoints.

Like this:

```python
_endpoints = {
    # ...
    'update_field': '/items/{item_id}/fields',
    # ...
}
```
Then this line should simply:

```python
url = self.build_url(self._endpoints.get('update_field').format(item_id=item_id))
```

O365/sharepoint.py
```diff
@@ -236,8 +240,41 @@ def get_list_columns(self):
...
+        
+        response = self.con.patch(url, update)
```
janscas: You should check if the response succeeded.

like so:

```python
if not response:
    # handle the return default when there is a Http error: For example return False to flag this operation as a fail.
```

O365/sharepoint.py
```diff
@@ -236,8 +240,41 @@ def get_list_columns(self):
...
+        
+        data = response.json()
```
janscas: data is not used here

O365/sharepoint.py
```diff
@@ -236,8 +240,41 @@ def get_list_columns(self):
...
+        
+        return self.get_item_by_id(item_id)
```
janscas: This does another api call.

You can avoid this by updating the inner properties that the user has updated.

Update method calls should return True/False flagging the success.

The changes are implicit when the call succeeds.

O365/sharepoint.py
```diff
@@ -236,8 +240,41 @@ def get_list_columns(self):
...
+        :param new_data: dictionary of {'col_name': col_value}
+                    
+        :rtype: SharepointListItem
+        """
+        
+        url = self.build_url(self._endpoints.get('get_items'))
+        
+        response = self.con.post(url, {'fields': new_data})
+        
+        data = response.json()
```
janscas: Again, check if the request failed:

```python
if not response:
     return None
```

O365/sharepoint.py
jtweeder: changed to return false if the update fails and true if it works.

O365/sharepoint.py
jtweeder: removed since it was not used.

O365/sharepoint.py
jtweeder: Removed the calling of the item from the API. 

Since the item does not have to be declared, and updates can be made with just the ID number, would it be more appropriate to treat the list like the mailbox, and the items like the message?  

This would move the update to the item class and the changes would be tracked within each item?  Was not sure what level the user should be able to update the list.  

Or am I missing something that could easily be done at the list level, like collecting the item ID, and fields to change and then pass that to the API in an update.

O365/sharepoint.py
jtweeder: Added success checking

jtweeder:
Thank you for the review.  I addressed the requested changes as best I could.  I reviewed the way message updates are handled and had a question on what level the update should be available to list items.  

Should the item have to be declared so that the properties are available and track able, or can the user just give an ID and the changes to the list item?  In mailbox I can see you can create, and delete a message, but not update a message.  You have to be in the message class to update the message.  

Does the list and list items get treated the same way?  Able to create, and delete a list item from the list class, but have to create a listitem to update it vs just passing the ID and updates to the listclass.  

I can see it both ways so I was not sure the best in terms of your vision and ease when the bulk feature is added in later.  

Thanks again for the opportunity to contribute, and thanks for this module.  Makes my life easier!

O365/sharepoint.py
janscas: Yep that's exactly what I was saying on the main review comment.
The update should happen in the item itself, and the item should be tracking the it's internal state to flag where there's been a change.

Then a simple item.update() (without any arguments) should "commit" only the properties changed to the servers.

janscas:
> Should the item have to be declared so that the properties are available and track able, or can the user just give an ID and the changes to the list item? In mailbox I can see you can create, and delete a message, but not update a message. You have to be in the message class to update the message.

That's correct, Calendar and Events work the same way.

> Does the list and list items get treated the same way? Able to create, and delete a list item from the list class, but have to create a listitem to update it vs just passing the ID and updates to the listclass.

It should work this way.

janscas:
But, as I said we can incrementally change this part of the api. So you can leave it like this for now if you don't want to enhance it more.

But at the end, this should be changed. Any ListItem should be able to update itself. And the list should not update a listitem.

jtweeder:
Now that i have it in my thick skull what you had originally said, I have working version with the update inside the listitem.  Let me finish cleaning it up and I will post the changes.

jtweeder:
I moved the update item to the list item class.  Let me know what changes you would like to see from that.  

I have also added a delete item to the list item class to round out the editing a list item functionality.

janscas:
That's great!

I'll merge it soon

PR #217, opened 2019-03-28
Use just resource_prefix if parent is a site

### LINKED ISSUE #214: main_resource is incorrect when using get_subsites
I have not been able to trace back where this is exactly happening to propose a solution, but here is the behavior.

First I created a sharepoint site.

```python
site = acct.sharepoint().get_site('root','/path/to_site')
```

This site conatins three subsites, and wanting to create a site instance from the first listed.
```python
sub_site = site.get_subsites()[0]
```

It seems to build the the site for the sub_site without issue until I try to interact with the API where I get a 400, Bad Request Resource not found.  Looking at the error it appeared that the url for the subsite was stacked  on top of the site information.  

Example Error:
```400 Client Error: Bad Request for url: https://graph.microsoft.com/v1.0/sites/<sharepointaddress>,<siteID>/sites/<sharepointaddress>,<subsiteID>```

Looking at the .main_resource of each of the objects I can see the repetition in the subsite.

```python
sub_site.main_resource
'sites/<sharepointaddress>,<SiteID>sites/<sharepointaddress>,<subsiteID>'
```

while the site.main_resource is as follows:
```'sites/<sharepointaddress>,<SiteID>'```

If I make a site object with a path directly to the site I want from the subsites it works, but then I cant use the listing of subsites to my advantage.

```sub_site =  acct.sharepoint().get_site('root','/path/to_site/subsite')```

Now the sub_site works as expected and has the following under sub_site.main_resource:
```'sites/<SharepointAddress>,<subsiteID>'```

I hope that explanation is clear enough for what I was experiencing, and feel free to ask any clarifying questions.

jtweeder:
When the site class is being built from a parent site, in the cases of get_subsites, the parent main_resource does not need to be used, only the resource_prefix.

Closes #214 

Fix works when building a site directly from Sharepoint class and also from subsite class.  Functionality is as expected for the resulting site class no matter how it was built with this fix.

### DIFF
--- a/O365/sharepoint.py
+++ b/O365/sharepoint.py
@@ -275,7 +275,8 @@ class Site(ApiComponent):
 
         # prefix with the current known site
         resource_prefix = 'sites/{site_id}'.format(site_id=self.object_id)
-        main_resource = '{}{}'.format(main_resource, resource_prefix)
+        main_resource = (resource_prefix if isinstance(parent,Site)
+                            else '{}{}'.format(main_resource, resource_prefix))
 
         super().__init__(
             protocol=parent.protocol if parent else kwargs.get('protocol'),


janscas:
Looks good! Thanks!

PR #224, opened 2019-04-08
Sharepoint Docs Draft

jtweeder:
Here is the first pass at the Sharepoint docs.  I have the HTML build from make html output, but I was not sure how you wanted to review the proposed documentation, and if there was already a process in place to build the new documentation.

After you review, just let me know if you want me to add in the entire HTML build output, parts of it, or just these .rst file changes if you had plans on making other updates to the docs.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -9,3 +9,4 @@ Detailed Usage
    usage/connection
    usage/account
    usage/mailbox
+   usage/sharepoint
--- a/docs/source/usage/sharepoint.rst
+++ b/docs/source/usage/sharepoint.rst
@@ -16,7 +16,7 @@ and :code:`.get_list_by_name('list_name')`.
 
 **Accessing Subsites**
 
-If a sharepoint site contains subsites they can be returned as a list of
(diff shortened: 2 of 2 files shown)


janscas:
I don't know how to roll this into the docs. This part was handled by @roycem90 

Let's wait a bit

stardust85:
I like that pull request (however I'm not an expert on writing docs - but who is?). It is better than no docs :D I'm working on file uploads and downloads as a user of this library now so if time permits (I don't want to promiss anything) I can improve it later by adding something about working with files

stardust85:
related issue: https://github.com/O365/python-o365/issues/409

stardust85:
so basically we will need to merge this first so we don't have a conflict...

janscas:
I look @GeethanadhP never review this.  I'll merge this.

stardust85:
Thanks for merging. Actually I managed to do my job with the library I used before (https://github.com/vgrem/Office365-REST-Python-Client) but in long term I would like to switch to yours because Office365-REST-Python-Client has inconvenient APIs

PR #231, opened 2019-04-16
fix bug for removing all categories from a message

jkamdjou:
`to_api_data()` will not serialize the `self.categories` property if you try to remove all categories from a message since `self.categories` evaluates to `False` for an empty list.

If `self.categories` has been modified, the `categories` string will be passed into `to_api_data()` from the `save_message()` function as the `restrict_keys` parameters.


janscas:
Good catch! thanks

janscas:
Unfortunately  this has introduced a bug when restrict_keys in fact is None.
Fixing it v.1.18

PR #261, opened 2019-06-13
Add ability to return list item field values

jtweeder:
This has helped us a lot in dealing with sharepoint lists, so I figured I would see about your thoughts on incorporating.

Ability to return the list item field value details when requesting all list items, or items by id number.  Expanded return is already handled by the list item fields attribute.

Let me know if you would like any changes.

### DIFF
--- a/O365/sharepoint.py
+++ b/O365/sharepoint.py
@@ -173,6 +173,7 @@ class SharepointList(ApiComponent):
     _endpoints = {
         'get_items': '/items',
         'get_item_by_id': '/items/{item_id}',
+        'expand': '?expand=fields',
         'get_list_columns': '/columns'
     }
     list_item_constructor = SharepointListItem
@@ -244,13 +245,16 @@ class SharepointList(ApiComponent):
         self.column_name_cw = {col.display_name: col.internal_name for
                                col in self.get_list_columns() if not col.read_only}
 
-    def get_items(self):
+    def get_items(self, expand=False):
         """ Returns a collection of Sharepoint Items
+        
+        :param expand (bool): Return the list items field details, default False
 
         :rtype: list[SharepointListItem]
         """
         url = self.build_url(self._endpoints.get('get_items'))
-
+        if expand:
+            url = ''.join([url, self._endpoints.get('expand')])
         response = self.con.get(url)
 
         if not response:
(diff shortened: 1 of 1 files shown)


janscas:
I'm not convinced, because the expand class is managed by the Query class.

Also note that you can expand more objects on "items" like:

- activities
- versions
- driveitem
- ...

See [Sharepoint ListItem Relationships](https://docs.microsoft.com/en-us/graph/api/resources/listitem?view=graph-rest-1.0#relationships)


Maybe the best way to do this is to add a query parameter that accepts a `Query` instance. Then apply this query to the `get_items` url. 

This works the same way messages or events (etc.) are queried.

Another approach could be to accept a parameter called `get_fields=False` that if True will set the query object that expands on "fields" and then return the appropriate values.

Also, you are not getting the result of the expand term. The expand on "fields" returns a "fieldValueSet"  that must be treated.

jtweeder:
Long time no contribution by me....I think I am going to start over on this approach, but wanted to get your thoughts.  

The functionality I am wanting to achieve is to return the items of a list and the values that they contain for specific columns, or all columns.  This would allow the use of data contained in a list to be extracted and used in other processes outside of sharepoint.

To clarify my understanding this can be done at the list level with the example at the bottom of [get list](https://docs.microsoft.com/en-us/graph/api/list-get?view=graph-rest-1.0&tabs=http) utilizing the `items(expand=fields(select=col1,col2)`.

So this would be built as a query instance similar to the way messages work?  (on my list to study further)  

You can also return list item values for a particular item using the `?expand=fields` using 
[get list item](https://docs.microsoft.com/en-us/graph/api/listitem-get?view=graph-rest-1.0&tabs=http) if you are at the list item level.

Also would this be handled as a query instance to accept the arguments you may want to pass to the `?expand=fields(select=.....`?

If I am pulling all the items in a list I would think pulling it from the list level versus iterating over all the list items would be preferred to minimize the number of api calls.

Thanks for your patience and understanding!

janscas:
@jtweeder When in a `Query` object you use **expand** and **select**. The query builds it like this:


This:
```python
q.expand('fields').select('col1', 'col2').as_params()
```
Outputs:
```python
{'$expand': 'fields($select=col2,col1)'}
```

PR #270, opened 2019-07-07
Support for webapp-style auth flow

toyg:
Added support for webapp-style auth flow.
Refactored some duplicate session-creating code.

This patch should allow people to do the full oauth dance as transparently as humanly possible, while still using this lib. Full example with a sample Flask app at https://gist.github.com/toyg/fa55d4b18f872c8f77dd58c9482acee2


toyg:
A practical example of how to use this flow is at https://gist.github.com/toyg/fa55d4b18f872c8f77dd58c9482acee2

TheSanjo:
Thanks ToyG, this is exactly what I have been waiting for.  I look forward to it being accepted.

janscas:
Hi thanks a lot for this pull request.
I see the point on some methods and not on others.
We try to maintain the library very unopinionated.

First, about this gist:

> https://gist.github.com/toyg/fa55d4b18f872c8f77dd58c9482acee2

- Connection does not accept a `redirect_uri` parameter. You don't have to pass it.

- You don't have to pass the scopes both to the `Connection` instance and the `get_authorization_webflow`. The scopes passed to the `Connection` instance are stored inside an attribute of the instance called `self.scopes`. See [get_authorization_url]( https://github.com/O365/python-o365/blob/master/O365/connection.py#L408).

- I understand this is an example but you are using a local variable to store 'state' between requests. This is a very bad practice since your server can span multiple instances and a single user can hit different server instances between requests and then any local variable stored will not be seen between different servers.

- As a recommendation don't store tokens in the file system (you are using the default TokenBackend) for the same reason stated before. Use a Custom TokenBackend that will communicate with any database you want.

- Also as a recommendation: Try to use the `Account` object. You can pass any params to the `Connection` in the initialization of the account and also access the connection methods by calling: `account.con` (alias) or `account.connection`. Such as: `account.con.request_token`.


About the Pull Request in general:

- I like the `_build_oauth2_session` method to create or re-create sessions. But it should be a public method. It should be used by the Connection class to create sessions and also by the users to re-build sessions as you did. In fact these alredy exists. It's called [get_session](https://github.com/O365/python-o365/blob/master/O365/connection.py#L478). But now requires a token stored so It can't be used prior to the authentication. Therefore `get_session` method should be changed to allow to build sessions without tokens.

- The `request_token` method should accept a 'state' and a 'redirect_uri' parameters to be able to re-build the session if needed. The scopes are not needed.

- I don't want to enforce a method to authenticate on webapps. These should up to the user. The library should do the minimal changes to allow an effective authentication flow for webapps. Therefore the methods `get_authorization_webflow` and `get_authorization_details` are not needed (see below). 
However I will like to say that:
    - To allow the re-creation of sessions the method `get_authorization_url` should return the state also. These will be changed in the next release (version 2.0).
    - The other params needed by the session are well known and therefore there's no need to store them along with the CSRF protection (the 'state').
    - I will add an example on how to handle authentication on web apps.


So at the end the web app flow should be:
1. First redirect the user to the authorization url. Store the state somewhere to recover it.
1. Receive the response in the redirect url (the callback). Then call `request_token` passing the authorization_url, the state and the redirect_uri.
1. At this point you have a token if the authentication succeeded.


I can't accept these pull request for what I mentioned but I will like to give credit to you about your ideas and changes so:

@toyg  **Can you create another pull request with the `get_session` and 'request_token` methods modified with your approach?**

Any comments?

janscas:
@toyg Tell me something as I want this to be released under version 2.0 which will be released in ten days aprox.

toyg:
The example was clearly a minimal, unopinionated snippet to show the mechanism. I personally find it confusing when too many concepts end up in an example, so I limited it to the bare minimum to show the flow - hence why I did not introduce a custom backend, which I obviously have in my own project.

I'm reworking the patch as requested, there is only one issue: using get_session() as a generic session builder means losing the RuntimeError about having no token. Is that acceptable?

janscas:
> there is only one issue: using get_session() as a generic session builder means losing the RuntimeError about having no token. Is that acceptable?

That's a very good question. Let me see...

toyg:
The other thing I was thinking of is to store the state as a field, rather than changing the signature of get_authorization_url to return it, so we don't break downstream.

janscas:
> there is only one issue: using get_session() as a generic session builder means losing the RuntimeError about having no token. Is that acceptable?

We can add an optional parameter to `get_session` called `check_token=False` (defaults to false) which will check for the presence of an auth token an raise the exception if not.

Then we change the internal `get_session` method calls which requires the token with `check_token=True`

janscas:
> The other thing I was thinking of is to store the state as a field, rather than changing the signature of get_authorization_url to return it, so we don't break downstream.

I also don't like to change the signature of method calls.. but saving it as an attribute of the class is kind of a hack... The user will have to retrieve it anyways to store it somewhere.

I prefer to change the method signature in this case... It was wrong from the beginning.

janscas:
Just as a recap:

I think this is the complete list of changes to allow web app authentication support (tell me if you need help on anything):

- [ ] Change `get_session` signature:
    - Add the following optional parameters:
        - `redirect_uri=None`
        - `state=None`
        - `load_token=False` (instead of check_token) : if True try to load the token from the backend. Raise RuntimeError if token is not present.
    - There is no need to add the client_id (credentials are always required and stored inside the auth property)
    - There is no need to add the scopes parameter. Those are only needed on the `get_authorization_url` step.
-  [ ] Change the `request_token` signature:
    - Add the following parameters:
        - `state=None`: to be able to rebuild the session if not present.
        - `redirect_uri=None`: to be able to rebuild the session if not present.
    - Rebuild the session with `get_session` if needed (if `self.session` is None).
- [ ] Return the `state`from the `get_authorization_url` method.
- [ ] Change the internal calls to the `get_session` method:
    - Modify `refresh_token` method. Set `self.session` if is None
    - Modify `oauth_request` method. Set `self.session` if is None
    - `get_authorization_url` is creating a session. Change it so it creates the session with  `get_session`
    - `request_token` must create a session if `self.session` is None. 
- [ ] Change the method `oauth_authentication_flow` on connection.py to gather the state from  `get_authorization_url`.

Also note that I've committed a number of significant changes in the previous days (and today). For example I've removed the 'token paths' related code from connection.py that had deprecation warnings on.

**Thanks a lot for all your help and for figuring out what was wrong.**

toyg:
Right, #278 should have it all.

PR #278, opened 2019-07-15
Changes required for web-flow authentication:

toyg:
- get_authorization_url() returns a (url, state) tuple
- get_session() takes additional parameters to (re)create sessions
- refactored internals to use get_session(), ensuring consistency

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -367,51 +367,45 @@ class Connection:
         :rtype: str
         """
 
-        client_id, client_secret = self.auth
-
-        if requested_scopes:
-            scopes = requested_scopes
-        elif self.scopes is not None:
-            scopes = self.scopes
-        else:
+        scopes = requested_scopes or self.scopes
+        if not scopes:
             raise ValueError('Must provide at least one scope')
 
-        self.session = oauth = OAuth2Session(client_id=client_id,
-                                             redirect_uri=redirect_uri,
-                                             scope=scopes)
-        self.session.proxies = self.proxy
-        if self.request_retries:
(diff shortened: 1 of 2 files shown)


janscas:
Pasting this from #270 just to have it all here:
<hr>

Just as a recap:

I think this is the complete list of changes to allow web app authentication support (tell me if you need help on anything):

- [ ] Change `get_session` signature:
    - Add the following optional parameters:
        - `redirect_uri=None`
        - `state=None`
        - `load_token=False` (instead of check_token) : if True try to load the token from the backend. Raise RuntimeError if token is not present.
    - There is no need to add the client_id (credentials are always required and stored inside the auth property)
    - There is no need to add the scopes parameter. Those are only needed on the `get_authorization_url` step.
-  [ ] Change the `request_token` signature:
    - Add the following parameters:
        - `state=None`: to be able to rebuild the session if not present.
        - `redirect_uri=None`: to be able to rebuild the session if not present.
    - Rebuild the session with `get_session` if needed (if `self.session` is None).
- [ ] Return the `state`from the `get_authorization_url` method.
- [ ] Change the internal calls to the `get_session` method:
    - Modify `refresh_token` method. Set `self.session` if is None
    - Modify `oauth_request` method. Set `self.session` if is None
    - `get_authorization_url` is creating a session. Change it so it creates the session with  `get_session`
    - `request_token` must create a session if `self.session` is None. 
- [ ] Change the method `oauth_authentication_flow` on connection.py to gather the state from  `get_authorization_url`.

janscas:
This is now on the latest release v2.0

PR #296, opened 2019-08-28
Add Drive.get_item_by_path(item_path)

cg6:
Allows for getting a specific file or folder from a drive or document library using the syntax:
`Drive.get_item_by_path('/path/to/item')`


janscas:
thanks

PR #299, opened 2019-08-31
Fix "keyerror" issue in beautifulsoup while checking attribute.

shyamdsundar:
Found this bug while parsing mails. Just a minor change required in the way the attributes are queried.


janscas:
Thanks!

PR #300, opened 2019-09-03
Set chunk_size to a valid value ('auto' is not a valid value in reque…

teamoo:
…sts)

If `chunk_size` is set to `'auto'` and the file is bigger than the chunk size, the method fails because the value `'auto'` is not of type `int` or `None`.

see [Response.iter_content()](https://2.python-requests.org/en/master/api/#requests.Response.iter_content)


janscas:
Thanks!

PR #301, opened 2019-09-04
Fix error when flag does not have start and due date

shyamdsundar:
This was throwing error when it was parsing flagged messages with no start and due date.


janscas:
Thanks

PR #307, opened 2019-09-16
Fix Table Delete Row method to match alternative docs

livejake:
See the following thread: https://github.com/microsoftgraph/microsoft-graph-docs/issues/4818
The current table row deletion doesn't work and returns the following "Error Message: Resource not found for the segment 'delete'."

The API method here does work: 
https://docs.microsoft.com/en-us/graph/api/resources/excel?view=graph-rest-beta#delete-table-row
I've updated the code and it now works.


janscas:
Great!

PR #309, opened 2019-09-17
Update table delete to match alternate in docs

livejake:
Similar to my fix to delete rows, I have another fix for deleting tables 
See: https://github.com/microsoftgraph/microsoft-graph-docs/issues/4188
Using the Post for deleting doesn't work (POST /workbook/tables/{id|name}/delete
) but using the Delete method (DELETE /workbook/worksheets/{id|name}/tables/{id|name}) does work


janscas:
Great thanks!

Issue #298, opened 2019-08-30
Different attributes for "user" and "contact" ressource are not properly supported


teamoo:
A [user](https://docs.microsoft.com/de-de/graph/api/resources/user?view=graph-rest-1.0#properties) resource has different attributes that a [contact](https://docs.microsoft.com/de-de/graph/api/resources/contact?view=graph-rest-1.0#properties) resource. When using the Users API (Global Address List), the Object Model should support slightly different attributes, e.g. a `Contact` has a `businessAddress ` property, whereas a `User` object does not have a `businessAddress` property, but a `streetAddress`,`postalCode`,`state`, etc.

I would consider this a feature request.

teamoo:
My idea would be: instead of creating all the getters and setters manually, why don't we do something like this:

In the constructor
```
        # store all properties
        self.__cloud_data__ = cloud_data
```

And define a new method (see [Customizing attribute access)](https://docs.python.org/3/reference/datamodel.html#customizing-attribute-access)        
```
    #dynamic getter for all other properties that are in cloud_data
    def __getattr__(self, name):
        try:
            return self.__cloud_data__[name]
        except KeyError:
            msg = "'{0}' object has no attribute '{1}'"
            raise AttributeError(msg.format(type(self).__name__, name))
```

teamoo:
I created a comparison of the attributes, maybe this helps when looking at this feature request.

User Ressource Attributes | Contract Ressource Attributes
-- | --
  | assistantName
aboutMe |  
accountEnabled |  
ageGroup |  
assignedLicenses |  
assignedPlans |  
birthday | birthday
  | businessAddress
  | businessHomePage
businessPhones | businessPhones
city |  
  | categories
  | changeKey
  | children
companyName | companyName
consentProvidedForMinor |  
country |  
createdDateTime | createdDateTime
department | department
displayName | displayName
employeeId |  
faxNumber |  
givenName | givenName
hireDate |  
id |  
mail* | emailAddresses*
  | fileAs
  | generation
  | homeAddress
  | homePhones
  | id
imAddresses | imAddresses
interests |  
  | initials
isResourceAccount |  
jobTitle | jobTitle
legalAgeGroupClassification |  
  | lastModifiedDateTime
licenseAssignmentStates |  
  | manager
  | middleName
mailboxSettings |  
mailNickname |  
mobilePhone | mobilePhone
mySite |  
  | nickName
officeLocation | officeLocation
onPremisesDistinguishedName |  
[...]

PR #312, opened 2019-09-18
added some properties of user resource

teamoo:
Just some properties that are important to me that are not on the contact resource. Change is not breaking anything for what I see.


janscas:
See #298

PR #313, opened 2019-09-19
Fixed assignment in scope (none) variable that throwed error and remo…

dhmaste:
Fixed assignment in scope variable which wasn't initialized and throwed error.
Also removed unnecessary requested_scopes when auth_flow_type == 'credentials' because the value is already stored in self.con.scopes

### DIFF
--- a/O365/account.py
+++ b/O365/account.py
@@ -36,8 +36,10 @@ class Account(object):
         elif auth_flow_type == 'credentials':
             # for client credential grant flow solely:
             # append the default scope if it's not provided
-            if not scopes:
-                kwargs['scopes'] = self.protocol.prefix_scope('.default')
+            if scopes:
+                kwargs['scopes'] = self.protocol.get_scopes_for(scopes)
+                if '.default' not in scopes:
+                    kwargs['scopes'].append(self.protocol.prefix_scope('.default'))
 
             # set main_resource to blank when it's the 'ME' resource
             if self.protocol.default_resource == ME_RESOURCE:
@@ -109,6 +111,16 @@ class Account(object):
                 return False
 
         elif self.con.auth_flow_type == 'credentials':
+            if scopes is not None:
+                if self.con.scopes is not None:
+                    raise RuntimeError('The scopes must be set either at the Account instantiation or on the account.authenticate method.')
+                else:
+                    self.con.scopes = self.protocol.get_scopes_for(scopes)
+                    if '.default' not in scopes:
+                        self.con.scopes.append(self.protocol.prefix_scope('.default'))
+            else:
+                if self.con.scopes is None:
+                    raise ValueError('The scopes are not set. Define the scopes requested.')
             return self.con.request_token(None)
         else:
             raise ValueError('Connection "auth_flow_type" must be either "authorization" or "credentials"')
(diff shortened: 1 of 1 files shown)


O365/account.py
```diff
@@ -110,7 +109,7 @@ def authenticate(self, *, scopes=None, **kwargs):
                 return False
 
         elif self.con.auth_flow_type == 'credentials':
-            return self.con.request_token(None, requested_scopes=scopes)
+            return self.con.request_token(None)
```
janscas: This is wrong. The user can pass the scopes to the authenticate method and so those must be passed to the request_token method.

dhmaste: In the case where auth_flow_type='authentication' the authenticate function permits only one assignment on the scopes. Either during the initialization of the Account or during the authenticate. That is why i thought that since for auth_flow_type='credentials', we assign the '.default' scope during Account initialization, it would not be valid to reassign the scope in the authentication.

janscas: ok, what I can do is change the implementation to require the scopes to be passed into the account init in case auth_flow_type is credentials (as this is needed for the auto-refresh to work).

Then we can do what you did.

dhmaste: If the '.default' scope is not needed then a similar approach to the auth_flow_type='authentication' could be implemented and require from the user to set the scopes in either init or authenticate.

janscas: The .default is needed, but the user can also pass another scopes such as 'offline_access'.

dhmaste:
Added more logic as we discussed. It follows the logic for the authentication auth_type_flow but also checks and adds '.default' if needed.

O365/account.py
```diff
@@ -36,8 +36,10 @@ def __init__(self, credentials, *, protocol=None, main_resource=None, **kwargs):
         elif auth_flow_type == 'credentials':
             # for client credential grant flow solely:
             # append the default scope if it's not provided
-            if not scopes:
-                kwargs['scopes'] = self.protocol.prefix_scope('.default')
+            if scopes:
```
janscas: You need the `if not scopes` to set the '.default' scope.

O365/account.py
```diff
@@ -36,8 +36,10 @@ def __init__(self, credentials, *, protocol=None, main_resource=None, **kwargs):
...
+                kwargs['scopes'] = self.protocol.get_scopes_for(scopes)
+                if '.default' not in scopes:
```
janscas: You can't rely on this. What if the user passed 'https://graph.microsoft.com/.default' to scopes?
Will then will add the .default scope twice.

O365/account.py
```diff
@@ -109,6 +111,16 @@ def authenticate(self, *, scopes=None, **kwargs):
                 return False
 
         elif self.con.auth_flow_type == 'credentials':
+            if scopes is not None:
+                if self.con.scopes is not None:
+                    raise RuntimeError('The scopes must be set either at the Account instantiation or on the account.authenticate method.')
+                else:
+                    self.con.scopes = self.protocol.get_scopes_for(scopes)
+                    if '.default' not in scopes:
```
janscas: Already said:

You can't rely on this. What if the user passed 'https://graph.microsoft.com/.default' to scopes?
Will then will add the .default scope twice.

Issue #332, opened 2019-10-16
get_message throws error with default parameters


NielsDebrier:
Hello,

While using this library we came a cross the following bug.

When using the method get_message from mailbox.py we get an error: 'NoneType' object has no attribute 'has_expands' when giving a valid object_id and leaving the rest on their default parameters.

When expecting the code I encountered the bug on line: https://github.com/O365/python-o365/blob/master/O365/mailbox.py#L151
It seems that there are missing parentheses here.
I think it should be: `query and (query.has_selects or query.has_expands):`

This is what happens:
first it evaluates `query and query.has_selects` this returns False because query is None (default). Next it tries to evaluate `False or query.has_expands` but has expands doesn't exist on None. So it throws an error.

if I have time I will try to make a pull request and link it here.

janscas:
Fixed in PR

PR #333, opened 2019-10-16
fixed missing parentheses which throws an error when using default parameters

NielsDebrier:
fixed missing parentheses which throws an error when using default parameters. See github issue #332


janscas:
Thanks!

PR #344, opened 2019-10-31
fix download 'AttributeError: __enter__' bug

tino-247:
@janscas FYI


janscas:
The use of a context manager is on purpose. If the stream param is True, then the response must be used as context manager so it can clean the stream nicely.

janscas:
Instead I'm going to look for the minimum requests version that has this utility and set it as a requirement.

PR #345, opened 2019-11-04
Implement alternative Protocol for financials API

teamoo:
See [Endpoints for the APIs for Dynamics 365 Business Central On-Premises and Online](https://docs.microsoft.com/en-us/dynamics-nav/api-reference/v1.0/endpoints-apis-for-dynamics) for additional details. This protocol can be used as an alternate protocol when using the financials endpoint.


janscas:
Thanks

PR #346, opened 2019-11-04
Bugfix

teamoo:
- parameter can be email or user id, so it should not be named
- the build_url call provides an unnamed parameter. This led to the following error: 

```
File "/home/xxx/.local/lib/python3.7/site-packages/O365/directory.py", line 261, in get_user
    url = self.build_url(self._endpoints.get('get_user').format(user))
KeyError: 'email'
```

### DIFF
--- a/O365/directory.py
+++ b/O365/directory.py
@@ -189,7 +189,7 @@ class User(ApiComponent):
 class Directory(ApiComponent):
 
     _endpoints = {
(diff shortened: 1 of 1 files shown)


teamoo:
Sorry, I just saw that you already fixed the bug in the next commit related to the profile photo but did not push it to pypi yet.

PR #352, opened 2019-11-07
Financials API initial commit

teamoo:
This file introduces the financials API endpoint. It is currently only available on the Graph API beta version or via the BusinessCentral API endpoint.

One additional dependency is created by this module (`inflection`).

Many endpoints are currently only available on the business central API endpoint. These are not yet fully implemented, whereas all endpoints accessible via Graph API are already fully implemented, but not all CRUD operations are yet tested for all entities.

Navigation properties are always loaded in the current implementation.

The object can be called like this:

```
scopes = ['Financials.ReadWrite.All']

credentials = (client_id, client_secret)

protocol = MSGraphProtocol(api_version='beta')
credentials = (client_id, client_secret)

account = Account(credentials, protocol=protocol)

if not account.is_authenticated:
    account.authenticate(scopes=scopes)

financials = Financials(parent=account)
company = financials.get_companies()
```

Currently, you cannot grant application access, only delegated access.
If using the BusinessCentral API endpoint, make sure to grant access not for "Microsoft Graph", but for "Dynamics 365 Business Central".

The `Company` object contains access to most objects as they are company-specific.

### DIFF
--- a/O365/financials.py
+++ b/O365/financials.py
@@ -181,7 +181,7 @@ class FinancialsApiComponent(ApiComponent):
 
         response = method(url, data=data)
 
-        if not response.status_code == 201:
+        if not response.status_code.ok:
             return False
 
         object_created = response.json()
--- a/O365/financials.py
+++ b/O365/financials.py
@@ -34,7 +34,7 @@ class FinancialsApiComponent(ApiComponent):
         if kwargs.pop('main_resource', None):
             main_resource = kwargs.pop('main_resource', None)
         elif parent:
-            if type(parent.protocol) == MSBusinessCentral365Protocol:
+            if isinstance(parent.protocol,MSBusinessCentral365Protocol):
                 main_resource = ('/companies/{id}').format(id=getattr(parent, 'id', None))
             else:
                 main_resource = ('financials/companies/{id}').format(id=getattr(parent, 'id', None))
@@ -55,7 +55,7 @@ class FinancialsApiComponent(ApiComponent):
         :return: final url
         :rtype: str
         """
-        if type(self.protocol) == MSBusinessCentral365Protocol:
+        if isinstance(self.protocol,MSBusinessCentral365Protocol):
             endpoint = replacenth(endpoint, "/", ")/", 2)
             endpoint = replacenth(endpoint, "/", "(", 1)
 
@@ -3984,7 +3984,7 @@ class Financials(ApiComponent):
         self.con = parent.con if parent else con
 
         # Choose the main_resource passed in kwargs over parent main_resource
-        if type(parent.protocol) == MSBusinessCentral365Protocol:
+        if isinstance(parent.protocol,MSBusinessCentral365Protocol):
             main_resource = ''
         else:
             main_resource = 'financials'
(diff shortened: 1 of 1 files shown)


janscas:
This is absolutely great @teamoo!

Let me go through the code and I'll report back

teamoo:
Sure, feel free to change things. I took a slightly different approach than you. In my opinion, you have a lot of boilerplate in the current classes, a lot of repetition in the code...event thought about using data classes (python>=3.7).

O365/financials.py
```diff
@@ -0,0 +1,4048 @@
...
+            return True  # there's nothing to update
+
+        url = self.build_url("/" + type(self)._ENTITY)
+
+        method = self.con.post
+        data = self.to_api_data(restrict_keys=self._track_changes)
+
+        response = method(url, data=data)
+
+        if not response.status_code == 201:
```
janscas: This in my experience with Microsoft Apis will sometimes fail. There are plenty of methods in the graph docs that should return 201 on created events, but returns other status codes. At the beginning of this library I was comparing to the exact status code on each method and I changed this to simply `if not reponse:` or if you prefer `if not response.ok` (both are the same).

O365/financials.py
```diff
@@ -0,0 +1,4048 @@
...
+        return self.__unit_of_measure
+
+    def update(self):
+        return super()._update()
+
+    def delete(self):
+        return super()._delete()
+
+    def create(self):
+        return super()._create()
```
janscas: Those are redundant? If they do nothing they can go away

teamoo: I am not sure I got your point. The methods create/update/delete are implemented in the parent class `FinancialsApiComponent` because I did not want to create redundancy by implementing the methods in every sub-class. The methods are hidden (_), because according to the documentation, you cannot CRUD on every object. For entities, that do not allow create/update/delete, the methods were not added.

O365/financials.py
teamoo: Done. Thanks for the tip.

O365/financials.py
```diff
@@ -0,0 +1,4048 @@
...
+        # internal to know which properties need to be updated on the server
+        self._track_changes = TrackerSet(casing=cc)
+
+        cloud_data = kwargs.get(self._cloud_data_key, {})
+
+        # Choose the main_resource passed in kwargs over parent main_resource
+        if kwargs.pop('main_resource', None):
+            main_resource = kwargs.pop('main_resource', None)
+        elif parent:
+            if type(parent.protocol) == MSBusinessCentral365Protocol:
```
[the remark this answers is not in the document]
teamoo: Done. Still using type() for company in two cases because in that cases I think it is safe and more accurate.

O365/financials.py
janscas: ok ok its create with underscore (_create). nevermind

O365/financials.py
```diff
@@ -181,7 +181,7 @@ def _create(self):
 
         response = method(url, data=data)
 
-        if not response.status_code == 201:
+        if not response.status_code.ok:
```
janscas: It's `response.ok`

janscas:
hi @teamoo can you change the code about the inflection dependency? There's any reason to maintain this dependency?


This way I can merge the PR.

teamoo:
Hi @janscas,

yes sure. Sorry I was busy with projects, will do that next week.

teamoo:
Sorry again, Deadline on a client project and time off next week. I will Update it in Deventer.

jeroenhabets:
@teamoo it's been rather quiet on your O365 Financials API pull request. Any change you'd be able to wrap it up soon?

teamoo:
Hi @jeroenhabets ,

sorry, no chance to update it at the moment. If it is a problem for you guys, feel free to discard it.

PR #356, opened 2019-11-10
Atomic file operations

riptusk331:
This should make sure that only one instance at a time can access a token file. See my comments in our issue discussion, as well as documentation in the methods. Tried to be as detailed as possible with logging to show how it works.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import inspect
 import os
 import time
 
@@ -534,7 +535,7 @@ class Connection:
 
         if load_token:
             # gets a fresh token from the store
-            token = self.token_backend.get_token()
+            token = self.token_backend.load_token()
             if token is None:
                 raise RuntimeError('No auth token found. Authentication Flow needed')
 
@@ -676,10 +677,13 @@ class Connection:
                                 f'retrying {_ - 1} more times.')
                     time.sleep(2)
                     log.debug('Waking up and rechecking token file for update'
-                              'from other instance...')
-                    self.token_backend.load_token()
+                              ' from other instance...')
+                    self.token_backend.token = self.token_backend.load_token()
             else:
-                self.token_backend.fs_wait = False
+                # don't change token state back from wait if we found the token
+                # was refreshed by another instance. this will be our signal
+                # to the calling method that we need to use a new session
+                self.session = self.get_session(load_token=True)
                 log.info('Token was refreshed by another instance...')
                 return True
         return False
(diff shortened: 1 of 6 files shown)


O365/utils/token.py
```diff
@@ -44,6 +44,26 @@ def expiration_datetime(self):
             expires_on = expires_on + dt.timedelta(days=90)
         return expires_on
 
+    @property
+    def access_expiry(self):
```
janscas: To maintain the same naming convention this should be called `access_expiration_datetime`.

O365/utils/token.py
```diff
@@ -44,6 +44,26 @@ def expiration_datetime(self):
...
+        """
+        Returns the token's access expiration datetime
+        :return datetime: The datetime the token's access expires
+        """
+        expires_at = self.get('expires_at')
+        if expires_at:
+            return dt.datetime.fromtimestamp(expires_at)
```
janscas: The `expiration_datetime` can now get the expiration value from this method and then add the 90 days if the token is long lived. This way we don't repeat code.

So you should substract the EXPIRES_ON_THRESHOLD. I think two minutes of threshold is maybe too much,.

O365/connection.py
```diff
@@ -534,7 +535,7 @@ def get_session(self, *, state=None,
 
         if load_token:
             # gets a fresh token from the store
-            token = self.token_backend.get_token()
+            token = self.token_backend.load_token()
```
janscas: I don't understand this. If you don't use `get_token` then the token backend 'token' property will remain unset. Therefore the state is inconsistent between the session and the token backend. I think this is fixed with the change I'll introduce later (the one I commented on in the Issue).

riptusk331: unintentional - i was debugging & inspecting the behavior of some of your changes and forgot to revert this back to `get_token`

O365/utils/token.py
riptusk331: agreed...removed repeat code & updated threshold to 1min

curious - what is the reason you have this threshold?

O365/utils/token.py
riptusk331: sure...done

O365/utils/token.py
janscas: Because if someone is checking the `is_expired` property with a refresh token enabled I want the user to be able to request the new refresh token before it expires. It just an extremely edge case, but I thought 2 minutes in a 90 day range is nothing.

riptusk331:
these are changes only to token.py. reverted out all changes to connection.

as we discussed, i will work up an locked file system subclass for the examples

Issue #388, opened 2020-01-12
Settings for connection json dumps encoder


dtorres-sf:
I am having an issue when trying to send date and datetime to an excel worksheet, since they are not json serializable. The problem is the json.dumps in the Connection:

https://github.com/O365/python-o365/blob/master/O365/connection.py#L658

If this would allow for a custom `cls` then it would work since we could set the encoder to use for dumping json before the request.

PR #389, opened 2020-01-12
Allow for custom JSON Encoder class

dtorres-sf:
Fix for #388 

This allows for a custom JSON encoder which is needed when using date and datetime objects. With this change you can do something like the following:

```
        account = Account(credentials, scopes=settings.O365_SCOPES, token_backend=token_backend, json_encoder=MyJSONEncoder)
```

Where `MyJSONEncoder` could do:

```
import json
import datetime
class ComplexEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, datetime.date):
            return obj.isoformat()
       return json.JSONEncoder.default(self, obj)
```

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -323,7 +323,7 @@ class Connection:
                  request_retries=3, token_backend=None,
                  tenant_id='common',
                  auth_flow_type='authorization',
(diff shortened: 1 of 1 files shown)


janscas:
Great

Issue #378, opened 2019-12-24
Microsoft Teams integration


vasjapavlovic:
When do you plan to integrate microsoft teams in the module?
I try to create a channel in a team. can you point out how to achive that?

Endpoint:
`https://graph.microsoft.com/v1.0/teams/{team-id}/channels`

Data:

```
{
  "displayName": "Architecture Discussion",
  "description": "This channel is where we debate all future architecture plans"
}
```

janscas:
There's currently no plans to add Microsoft Teams support.

matt43121:
Hi,

Are you accepting pull requests on this? I have some code based on your implementation of the planner resource if you are that may satisfy this need.

janscas:
Of course. It just needs to comply with the library code style and philosophy.

matt43121:
@janscas Just looking back at this. In order to expand teams integration, most of the base of MS Teams are on the groups/users endpoints. Are you wanting the users and group endpoints to be in unique files separate from the Teams code or not bothered?

janscas:
Hi @matt43121 
Thanks for asking

The [User ](https://github.com/O365/python-o365/blob/master/O365/directory.py#L12)object is already done (lacking some methods) in the directory.py file.

I think Groups should have their own file because is a separate functionality (but related to teams and more).

tomiszili:
Hi all!

Could anybody help me, i dind't find any example or documentation about this [problem](https://github.com/O365/python-o365/issues/429#issue-605500797)?

Thanks!

alejcas:
Teams was added later on

PR #410, opened 2020-03-18
Initial Microsoft Teams support

matt43121:
This is a start to adding Microsoft Teams support connected to issue #378. 

This adds basic support for creating channels, getting the logged in users teams, getting channel info and apps used within a team. I have not fleshed this out too much as most of the Teams endpoints are linked/rely on Groups so this may be something to look into in the future. 

As this is my first pull request I was not too sure about the library-style however as this is quite a basic change I was able to base the code heavily on planner.py.

### DIFF
--- a/O365/account.py
+++ b/O365/account.py
@@ -238,6 +238,16 @@ class Account:
 
         from .planner import Planner
         return Planner(parent=self, main_resource=resource)
+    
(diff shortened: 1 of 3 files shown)


janscas:
That is perfect!

Thanks!

PR #425, opened 2020-04-16
feat(attachment): add support to download item attachments as eml files

icamposrivera:


janscas:
This is already implemented in https://github.com/O365/python-o365/blob/master/O365/message.py#L55 (Save attachment as eml) and https://github.com/O365/python-o365/blob/master/O365/message.py#L1060 (Save message as eml)

icamposrivera:
Sorry Alejandro! I did not see the feature :'(

PR #433, opened 2020-04-29
get extra params by means of query in directory getuser method

icamposrivera:


icamposrivera:
origin is a dirty branch, sorry

PR #459, opened 2020-05-25
Public Client

JonPusey:
Added ability to request token without supplying client secret by setting auth_flow_type to 'public'


janscas:
There's a little room for improvement but I'll merge this.

JonPusey:
Thankyou. I have time to make improvements if you have time to suggest some.

janscas:
There's a subtle error in your code. 

Param "credentials" should be always a tuple. So when using public you should set the credentials as a 1 element tuple like:

`credentials = (app_id, )`

I'm committing a change with some code optimisation and this. I'm also updating the readme.

PR #460, opened 2020-05-26
Add enable\disable\force password change directory methods

i223t:
Add some methods to be able to enable\disable users or force to change the password. Not sure, is it a right class or I should use User?


janscas:
> Not sure, is it a right class or I should use User?

It's ok although this methods should also be available on the User class

janscas:
I've added a couple of comments to your code. They apply for all the methods.

Thanks!

PR #465, opened 2020-06-05
Improve convert_to_pdf integration

l-kotzur:
A small proposal to include a validation to convert_to_pdf:

- Raise ValueError if wrong file extension is given

- Make file extension validation on original file (`self.name`) and not on target file (`name`)

Especially, the last point gave me short headache since my target file name was obviously `name.pdf`, while `,pdf` is not included in the allowed extensions.

Hope it helps!

Best
Leander


janscas:
Great

PR #467, opened 2020-06-13
Fixed failed imports in tests. Added conversationIndex property to Me…

viralogic:
…ssage class.


O365/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .account import Account
 from .connection import Connection, Protocol, MSGraphProtocol, MSOffice365Protocol
 from .utils import FileSystemTokenBackend
+from .message import Message
```
janscas, line 11: Message should not be imported by default.
If you want to import Message, just:  `from O365.message import Message`

If the import fails in the tests, then the tests should change to import Message as said.

viralogic: Thanks for the feedback. Your requested changes have been made.

PR #468, opened 2020-06-18
Add upload and download stream option

ryanscovill:
I've added the ability to pass in io objects to write to/read from in the file upload and download, in addition to the filename.  This allows easier integration with the project I'm working on.


janscas:
good

PR #505, opened 2020-08-24
Conflict handling support for upload_file

gerkevgarderen:
Main change / new 'feature':
- Added an input argument that can be used to control the conflict handling when uploading a file in chunks.

Smaller changes:
 -  Fixed the parameter name `item_name` so the documentation fits the actual input argument
 -  Added one sentence to chunk_size so it's clear when it will be relevant.

The default now overwrites a file (despite MS docs saying it will fail), without raising errors which seems error-prone to me.
(see [MS docs](https://docs.microsoft.com/en-us/graph/api/driveitem-createuploadsession?view=graph-rest-1.0#create-an-upload-session) for the official documentation)

This option is sadly unavailable for non-resumable sessions as of now (not implemented in Graph API)

Not entirely sure on how you like to use naming/explanations at the moment so please take a look. 

Let me know if you have any questions/improvements!

### DIFF
--- a/O365/drive.py
+++ b/O365/drive.py
@@ -1210,19 +1210,24 @@ class Folder(DriveItem):
             return items
 
     def upload_file(self, item, item_name=None, chunk_size=DEFAULT_UPLOAD_CHUNK_SIZE,
-                    upload_in_chunks=False, stream=None, stream_size=None):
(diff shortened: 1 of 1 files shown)


janscas:
That's great. thanks!

PR #515, opened 2020-09-16
Update connection.py to pass kwargs to FileSystemTokenBackend

adamlogan73:
Need this to override file path/name of the token.


janscas:
Thanks!

PR #524, opened 2020-09-23
Tasks support

alxhslm:
I've added basic support for Outlook Tasks, based on the existing Calendar architecture. All of the basic task functionality offered by the MSOffice365 API is supported including:

- Addition or removal of tasks and folders
- Completing or uncompleting tasks
- Setting due dates

However, attachments and reminders are not yet supported. The MSGraph API for tasks [here](https://docs.microsoft.com/en-us/graph/api/resources/todo-overview) is still in beta so I haven't bothered to add support for this yet. I've successfully used this implementation to migrate all my tasks from Todoist over to Microsoft Todo.


janscas:
That's great work.

Could yo also add a proper section on the Readme?
Thanks!

PR #540, opened 2020-11-06
FIX: drive.py

nimishchaudhari:
Fixed: Crashes when trying to download contents from a folder that has an empty folder.


janscas:
thanks!

PR #544, opened 2020-11-19
Add AWS Secrets Manager Token Backend

e-gons:
I propose adding [AWS Secrets manager](https://aws.amazon.com/secrets-manager/) as a backend for token storage.


janscas:
That's great!

PR #547, opened 2020-11-28
Update calendar.py

Ombental:
The get method of attachment updating didn't have a relevant endpoint.

I don't know how to handle the attachments changing except for the built in method in the attachments class...

### DIFF
--- a/O365/calendar.py
+++ b/O365/calendar.py
@@ -961,6 +961,9 @@ class Event(ApiComponent, AttachableMixin, HandleRecipientsMixin):
         if self.__recurrence:
             data[cc('recurrence')] = self.__recurrence.to_api_data()
         
+        if self.has_attachments:
+            data[cc('attachments')] = self.__attachments.to_api_data()
+
         if 'attachments' in restrict_keys:
             self.attachments._update_attachments_to_cloud()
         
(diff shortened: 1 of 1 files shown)


O365/calendar.py
```diff
@@ -958,9 +961,13 @@ def to_api_data(self, restrict_keys=None):
         if self.__recurrence:
             data[cc('recurrence')] = self.__recurrence.to_api_data()
 
-        if self.has_attachments:
-            data[cc('attachments')] = self.__attachments.to_api_data()
-
+#         if self.has_attachments:
+#             data[cc('attachments')] = self.__attachments.to_api_data()
+#  Event object doesn't have __attachments attribute...
```
Ombental, line 966: useless comment

janscas: `Event` object does have the "__attachments" object. So this should be included (not commented out)

O365/calendar.py
```diff
@@ -958,9 +961,13 @@ def to_api_data(self, restrict_keys=None):
...
+        
+        if 'attachments' in restrict_keys:
+            self.attachments._update_attachments_to_cloud()
```
janscas, line 965: That's ok

janscas (commented):
I can't merge this as it will not include the attachments if there's an update because you commented a key piece of code. Can u revert it?

The other changes are ok and should fix the get attachment method.

janscas:
thanks!

PR #560, opened 2020-12-18
AWS tokens

jrmatchett:
I added token storage backends for Amazon Web Services S3 and Secrets Manager. Please note that this pull request makes significant modifications to the AWS Secrets Manager backend that was developed by [e-gons](https://github.com/e-gons) (issue #544 ). In particular, I removed the requirement to specify an `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`. In general, using those parameters to establish an AWS client should only be done in a local development environment (authentication in a production setting should be done via [Identity Access & Management](https://aws.amazon.com/iam/)). My AWS backends can still be utilized using an access key ID and secret when those values have been [setup using the AWS Command Line Interface](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html) (the `boto3` package automatically retrieves the values).


janscas:
Thanks!

PR #561, opened 2020-12-20
Feature/add teams presence

Squixx:
Adds functionality to monitor teams presence (useful for e.g having a smart light outside of your homeoffice change color when in a call) 

Requires the Presence.Read scope to be available

Let me know if I need to add some more docs


janscas:
Great!

PR #562, opened 2020-12-20
Add Task.importance and Task.is_starred properties

berezovskyi:
This PR adds a read-only `importance` property as per https://docs.microsoft.com/en-us/graph/api/resources/todotask?view=graph-rest-1.0#properties

Also, an experimentally established level "High" was found to match the MS To Do task when starred, and `is_starred` implements this heuristic.

### DIFF
--- a/O365/tasks.py
+++ b/O365/tasks.py
@@ -167,7 +167,7 @@ class Task(ApiComponent):
 
     @property
     def importance(self):
-        """ Task importance
+        """ Task importance (Low, Normal, High)
 
         :getter: Get importance level
         :type: str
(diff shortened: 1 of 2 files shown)


janscas:
thanks!

berezovskyi:
Thank you! I meant to clean up the PR before submitting for review but I must admit to my great shame what I was planning to do since. Please check what needs to be done before merging.

janscas:
Done!! thanks!

berezovskyi:
Ah, now it came back to me: I was thinking to make those properties writable but could not figure out how to do so from the first look. Anyway, that's for the future improvement. Thank you very much for maintaining the library!

PR #577, opened 2021-02-18
Fix scopes process of get_authorization_url

obsd:
The function `get_authorization_url` is passing scopes to `get_session` without process, so you get an error that there aren't such scopes.
You can see here how the scopes look like before the fix and compare them to the scopes that did work:
```
print(requested_scopes, self.scopes)
['basic', 'message_all'] ['https://graph.microsoft.com/Mail.ReadWrite', 'https://graph.microsoft.com/Mail.Send', 'offline_access', 'https://graph.microsoft.com/User.Read']
```
and after the fix:

```
protocol = MSGraphProtocol()
protocol.get_scopes_for(requested_scopes)
['https://graph.microsoft.com/Mail.ReadWrite', 'https://graph.microsoft.com/Mail.Send', 'offline_access', 'https://graph.microsoft.com/User.Read']
```

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -448,7 +448,12 @@ class Connection:
 
         redirect_uri = redirect_uri or self.oauth_redirect_url
 
-        scopes = requested_scopes or self.scopes
+        if requested_scopes:
+            protocol = MSGraphProtocol()
+            requested_scopes = protocol.get_scopes_for(requested_scopes)
+            scopes = requested_scopes
+        else:
+            scopes = self.scopes
         if not scopes:
             raise ValueError('Must provide at least one scope')
 


odedvisiblerisk:
@janscas can you review it?

O365/connection.py
```diff
@@ -448,7 +448,12 @@ def get_authorization_url(self, requested_scopes=None,
 
         redirect_uri = redirect_uri or self.oauth_redirect_url
 
-        scopes = requested_scopes or self.scopes
+        if requested_scopes:
+            protocol = MSGraphProtocol()
```
janscas, line 452: You can't assume here that the protocol in use is `MSGraphProtocol`.

janscas (commented):
Hi!
I know, but this is on purpose. The connection is "protocol" agnostic. It doesn't know the protocol used and you can't assume is `MsGraphProtocol`.

This is somehow a design decision made a long time ago.
For this reason, when not using the `Account` object, you should use a `Connection` and a `Protocol`.
Pass the scopes to the protocol `get_scopes_for` method and then pass the returned scopes to the `Connection` (or directly to the method `get_authorization_url`if you prefer).

Issue #487, opened 2020-07-23
request for an example to retrieve a sharepoint list into a table


InitialXW:
I am new to O365, i have searched the doc and the issues sections but could not get my code to read from a sharepoint list into a table-like structure for further processing, wonder if i am missing anything obvious here. the below is the code i have so far, any help is greatly appreciated, thanks:

resultlist = myspsite.get_list_by_name('my list name')
items = resultlist.get_items()
for item in items:
      print(item.data)

# then i get:
[List Item: https://mysp.sharepoint.com/sites/SITENAME/NAME/Lists/MYLISTItems/1289_.000, List Item: https://mysp.sharepoint.com/sites/SITENAME/NAME/Lists/MYLISTItems/1290_.000, ...]
what do i need to do from here to read the list/items into e.g. a pandas table?

jjr5330:
I faced the same issue. I believe you should be able to print item.fields instead of item.data and get the actual data returned. However, this feature seems to be broken as I couldn't get it to work either. It returns None values for every record me.

This is how I was able to build a dataframe with the SharePoint list data. I would really like feedback if anyone else has done this in an easier way because I think there should definitely be an easier way. Or if someone can find where the code is broken in SharePoint list .get_items() method and could mimic the get_item_by_id() method's fields output. It seems like the SharepointListItem class is not allowing for field values to be returned for some reason.

Note: I use pandas to create the dataframe if you are unfamiliar it is an awesome python tool for reading in data in a table like structure.

```
# use sharepoint site to get list by name
share_point_list = site.get_list_by_name('INSERT YOUR LIST NAME HERE')

# get share point list items
list_items = share_point_list.get_items()

# create a list of object_id values for all sharepoint list item rows
id_list = []

for item in list_items:
    id_list.append(item.object_id)

# create a list of share point list item field value dictionaries using the ids
# note you will use the share point list object we created earlier
list_row_values = []

for item_id in id_list:
    list_row = share_point_list.get_item_by_id(item_id)
    list_row_values.append(list_row.fields)

# create pandas dataframe from list_row_values
df = pd.DataFrame(list_row_values)

# save dataframe to csv if needed
df.to_csv(R"insert_file_path\sharepoint_list.csv", index=False)
```

InitialXW:
Thanks so much for the help! I echo your wonder on the use of ".field" and I got everything returned with "None' value too. This is exactly where i scratched my head and wondered if i missed anything obvioius.
[...]

Issue #436, opened 2020-04-30
sharepoint listitem expand=fields


techon:
Add support for the URL parameter ?expand=fields
SharepointList.get_items() 

Or is there another way to pull any additional fields?

janscas:
I'm not into this for the moment. If you want PR are welcome

techon:
It's only a few lines to add this to the code as a parameter for the
function, but I'm unable to push my code update due to permissions.

janscas:
You can fork and make a Pr without problems. I can't help you beyond that

techon:
I’ll handle it on my own… 


I was just trying to help you and your project is all.

kwollaston:
@techon, you would be helping if you created a PR like he mentioned - it is customary to fork repositories you don't own.. 

Anyways, I ran in to this issue as well and went ahead and submitted a PR for you @janscas. Please review/make changes as necessary. Wasn't sure how you wanted to handle an invalid field name (I used log.warning)

janscas:
@kwollaston thank you!

PR #581, opened 2021-02-21
Add support for list item fields

kwollaston:
* Added build_field_filter helper function that will add fields to the query based on their internal name or display name
* Added expand_fields argument to get_items and get_item_by_id methods, if you pass True it will return all fields, if you pass a list of field names, it will filter as such


kwollaston:
Related issues: #487 #436

janscas:
Great job, thanks

PR #582, opened 2021-02-22
Update TEST_SCOPES in protocol unit test with additional scopes

kwollaston:
* Update `TEST_SCOPES` in protocol unit test with additional scopes
* Used `if hasattr(self, 'attr'):` instead of `if self.attr:` as I was getting an `AttributeError`

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -857,7 +857,7 @@ class Connection:
         There is no guarantee that this method will be called by the garbage collection
         But this is not an issue because this connections will be automatically closed.
         """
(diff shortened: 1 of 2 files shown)


kwollaston:
The unit tests now pass, I would recommend moving the build to Github Actions.

janscas:
Hi @kwollaston the tests are a pending point on this project. Something was done a long time ago but nothing else. In fact the current tests are somewhat between wrong and null effect.

Anyway, merging this.

Thanks

Issue #584, opened 2021-02-24
Remove 3.5 from travis.yaml


cazador481:
The code uses f strings which do not supported until python 3.6.  So the python version 3.5 needs to be removed.

cazador481:
Looks like someone already submitted a patch to remove the fstrings.  closing

PR #585, opened 2021-02-24
issue #584: Remove python 3.5 from travis.

cazador481:
The code uses f-strings which aren't supported till 3.6  .

Along the way I also changed the minimum python version in the setup.py file to be python 3.6

Tested by setting up my own travis in github, and ran it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-- '3.5'
 - '3.6'
 install:
 - pip install -r requirements-dev.txt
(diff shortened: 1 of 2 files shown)


kwollaston:
Hi @cazador481, the only instance of f-strings being used is in `utils/token.py`.
Therefore, I think the proper fix here is to change the f strings to the backwards compatible format. `"x: {} y: {}".format(x, y)`

The Github readme states:
"Requirements: >= Python 3.4"

The PyPI package specifies 3.4 as the minimum Python version.

I will also quote this from @janscas on what I guess is the original package (pyo365):
"Python 3.4 is the minimum required... I was very tempted to just go for 3.6 and use f-strings. Those are fantastic!"

janscas:
Correct! @kwollaston already submit a PR with the fix. Thanks!

PR #590, opened 2021-03-02
Update connection.py

kwollaston:
I have found that I couldn't assign a single value tuple without a trailing comma. If credentials is provided as `(app_id)` it ends up being turned in to a str (~~even if you explicitly cast it as a tuple~~ edit: this actually splits the string in to a tuple of it's chars) thus failing the tuple instance check. If you use `(app_id,)` everything works.

I added a check that will ensure it is a single value tuple by converting string to tuple (using the necessary trailing comma).


janscas:
Good!

PR #598, opened 2021-03-17
Allow pulling of DriveItems external to tenant

### LINKED ISSUE #513: Pulling DriveItems external to tenant
It looks like the [get_shared_with_me()](https://github.com/O365/python-o365/blob/master/O365/drive.py#L1557) method in storage.py by default only pulls in DriveItems that are internal to the user's tenant. 

According to the [Microsoft docs,](https://docs.microsoft.com/en-us/graph/api/drive-sharedwithme?view=graph-rest-1.0&tabs=http) , adding `?allowexternal=true` to the GET request allows the call to pull in external DriveItems. I don't see that currently supported in any way (Correct me if I'm wrong on that point, I'm pretty sure that it works outside of the query functionality for instance).

Possible to add support for this? Happy to submit a PR if I can confirm that this isn't supported.

mcoirad-gmmb:
Resolves #513

### DIFF
--- a/O365/drive.py
+++ b/O365/drive.py
@@ -1580,7 +1580,7 @@ class Drive(ApiComponent):
             url = self.build_url(self._endpoints.get('shared_with_me_default'))
 
         # whether to include driveitems external to tenant
-        params = {"$allowexternal" : allow_external}
+        params = {"allowexternal" : allow_external}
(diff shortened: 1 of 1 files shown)


janscas:
thanks!

PR #618, opened 2021-04-29
Allow a dynamic consent process using functions

Perzan:
This paves the way for more elegant alternatives for attaining a token using a consent URL. Example implementations would not be limited to using a web browser, a pyperclip copy, or a web browser combined with a custom redirect_uri and a temporary HTTP server on localhost.
The default and current consent process function is O365.utils.consent.consent_input_token.


Perzan:
That built-in/default method could probably renamed to "consent_console".

janscas:
Good addition! But I’m not sure how this will handle two step authentication methods.

PR #623, opened 2021-05-12
Use batch argument even when limit is huge

### LINKED ISSUE #622: When limit is bigger than protocols' max then batch size is always overridden
When calling `Folder.get_messages` with a limit higher than the `protocol.max_top_value` (or `None`) than the batch size will be overridden: `batch = self.protocol.max_top_value`. That way we can't handle latency issues since the batch size can't be controlled.

Perhaps it can be changed so if a `batch` argument is given it will not be overridden, regardless of the `limit` argument's value.

BentalGuy:
If a query has a limit which is bigger than the protocol's max_top_value than it only makes sense to batch it, and even more let the programmer decide the batch size.

Hope it doesn't break anything but it seems to me that if a batch argument can work with `limit = max_top_value  - 1` than it should also work with  `limit = max_top_value  + 1` .

Resolves #622

### DIFF
--- a/O365/mailbox.py
+++ b/O365/mailbox.py
@@ -195,7 +195,7 @@ class Folder(ApiComponent):
             url = self.build_url(self._endpoints.get('folder_messages').format(
                 id=self.folder_id))
 
-        if limit is None or limit > self.protocol.max_top_value:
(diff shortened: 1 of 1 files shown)


BentalGuy:
Fixes #622

janscas:
thanks

PR #633, opened 2021-05-20
Create upload session for attachment in messages

Naxest:
API Graph has a limitation of 4MB for each attachment.

The way to solve it is to create an upload session and PUT chunks of bytes until complete the file -> https://docs.microsoft.com/en-us/graph/api/attachment-createuploadsession?view=graph-rest-1.0&tabs=javascript

This change add support to files greater than 4Mb, it automatically check if the file to upload has more than 4Mb and choose this mechanism. 
I've taken some similar code from drive.py.

### DIFF
--- a/O365/message.py
+++ b/O365/message.py
@@ -49,6 +49,8 @@ class MessageAttachments(BaseAttachments):
         'attachments': '/messages/{id}/attachments',
         'attachment': '/messages/{id}/attachments/{ida}',
         'get_mime': '/messages/{id}/attachments/{ida}/$value',
+        'create_upload_session': '/messages/{id}/attachments/createUploadSession'
+
     }
     _attachment_constructor = MessageAttachment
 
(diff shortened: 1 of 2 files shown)


janscas:
Hi @Naxest that’s a great addition.
Just to make sure, did u check the code from drive.py that uploads files?
There’s also a `CopyOperation` object there that informs about the upload progress.

Anyway I’ll merge this as it’s important but just wanted to provide you with the way we solved this for drive uploads. I’m not sure if these two “upload” methods share the same api construct.

Thanks

Naxest:
Hi @janscas, thank you for your quick answer.

I've checked the code, but the `CopyOperation` is only used in `copy()` method, I'm not sure if it would work for `_update_attachments_to_cloud`. 
The implementation is based on `upload_file` from drive.py, it has different api construct for two reasons:
- I didn't want to change the structure and method syntax of the original.
- The upload session interactions with Graph is not exactly the same as drive.

Many thanks

janscas:
Great!

Naxest:
> @Naxest do you have any example how to use your pull request? You added endpoint but there's no implemented method for that. I'd like to send message with big attachment.

You don't need to use a special method, there is a check in `_update_attachments_to_cloud`, in case the file size exceeds a threshold, the create_upload_session will be used. So it is a feature and you don't need to worry about the file size.

PR #666, opened 2021-08-08
message: Check if attachments exists lazily

deanc12:
I ran some performance tests on this library and found a performance improvement that's easy to fix.
Instead of checking if there are attachments for every message I changed the `has_attachment` property to be 'lazy'.

Thanks guys, keep up the great work!


deanc12:
Hey, can someone review this please?

deanc12:
?

janscas:
Great work @deanc12 

Sorry for the late response... I was on vacation

Thanks

PR #682, opened 2021-09-10
LegacyApplicationClient support

Addarkar:
Dear, 
Would you please integrate LegacyApplicationClient authentication into this famous library?
THX a lot!


janscas:
what's legacy?

Can you adapt the readme?

PR #692, opened 2021-09-29
Use upload session when attachment is larger then 3MB

DaniEzzeddine:
Just had a bug when attachment with size 3.8MB was throwing error

From official docs https://docs.microsoft.com/en-us/graph/outlook-large-attachments?tabs=http:
> - If the file size is under 3 MB, do a single POST on the attachments navigation property of the Outlook item; see how to do this for a message or for an event. The successful POST response includes the ID of the file attachment.
> - If the file size is between 3MB and 150MB, create an upload session, and iteratively use PUT to upload ranges of bytes of the file until you have uploaded the entire file. A header in the final successful PUT response includes a URL with the attachment ID.

### DIFF
--- a/O365/utils/attachment.py
+++ b/O365/utils/attachment.py
@@ -7,8 +7,8 @@ from .utils import ApiComponent
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


janscas:
Thank!

PR #717, opened 2021-11-22
calendar.py bugfix: AttributeError: 'Attendee' object has no attribute 'email'

Gregorek85:
I have come accross small issue when trying to check if email is in the list of current attendees of event:

```
return item in {attendee.email for attendee in self.__attendees}
AttributeError: 'Attendee' object has no attribute 'email'.
```

When I checked the code "Attendee" indeed does not have attribute "email". 
It has attribute address instead. This PR provides a quick & easy fix for that.


janscas:
Thanks!

PR #761, opened 2022-04-07
Extended directory class to perform manager and directReports lookups.

dionm:
These changes add two new functions into `directory.py` class, to perform lookups for both `manager` and `directReports` endpoints.

**Manager**
`https://graph.microsoft.com/v1.0/users/{mail -or- id}/manager`

**Direct Reports**
`https://graph.microsoft.com/v1.0/users/{mail -or- id}/directReports`


janscas:
Thanks!

PR #767, opened 2022-04-16
Group class enhancements

Krukosz:
- added retrieving group by its mail field
- removed retrievieng group by its nickname field (it hadn't worked at all)
- added getting group members with "recursive" flag -> if group member is another group then get members of child group too


janscas:
Thanks!

PR #773, opened 2022-04-27
teams class fixes

Krukosz:
- when accessing channels with api, must provide team_id, it wasn't provided in endpoints addresses (tested only with Channels),i don't have permissions to Chats, so i cannot test it.
- in utils.py - commented out lines from 356, because this function replaces "channel" endpoint with "users", for my purposes it works, but i don't know what idea was of this specific elif

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,6 +727,20 @@ message.attachments.add('george_best_quotes.txt')
 message.save_draft()  # save the message on the cloud as a draft in the drafts folder
 ```
 
+##### Sending message with attachments bigger than 4MB
+
+MS Graph has attachment limitation to ~4MB when sending emails. To send message with bigger attachments just execute `save_draft()` method before sending, so above code would be like:
+
+```python
+message = mailbox.new_message()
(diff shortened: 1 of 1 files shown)


hcallen:
This might break the ability to to access channel messages when starting from the Teams class. Seems like the same can be accomplished by adding a 'get_channel_by_id' method to the Teams class.

Krukosz:
I wanted simply to send message in one of my channels. I looped through channels within team, catched appropiate one and then i had API error, that request is invalid. 
Request url was `/users/channels/{channelid}/{operation} `
instead of `/teams/{teamid}/channels/{channelid}/{operation}`

hcallen:
I see what you're saying.

I think the quick and easy fix is to update the if/elif chain in utils.py to account for the '@' in the resource.

hcallen:
I think I fixed the resource parsing in [776](https://github.com/O365/python-o365/pull/776) and wrote tests to make sure the module was working as intended.

PR #776, opened 2022-04-30
fix teams resource parsing

hcallen:
I added a few methods to the teams module and streamlined the resource parsing a bit.


janscas:
Great

PR #778, opened 2022-05-11
Support large mail attachments from memory

sebastiant:
Current functionality for upload of attachments greater than 3MB in size is only supported for attachments written to disk (for in-memory attachments it throws an `AttributeError`). With this change, support is added for in-memory attachments, stored in the `content` attribute


janscas:
Thanks!

sebastiant:
@janscas just discovered an introduced bug in this patch. I'll publish a PR later today, but don't publish a release including this code before that!

janscas:
Ok!

PR #791, opened 2022-06-12
add pagination to get_all_chats

jhoult:
When calling get_all_chats it does not paginate like the other get_all_XXX functions.  I copied the code from those and modified it for the correct endpoint to get all chats with pagination


janscas:
Thanks

Issue #6, opened 2015-04-22
Unit tests


Narcolapser:
Unit tests for all the modules.

Narcolapser:
completed with c4f68f362ee5ed8aed46132d011f4dc5470808a2

PR #793, opened 2022-06-16
Add tests for Message class

strokirk:
Hello,

This PR adds basic smoke tests for the `Message` class, including rudimentary snapshots of expected API network calls. 

This library has been very useful to me, and I'd like to help contribute where I can. 😃  While basic tests for were added with https://github.com/O365/python-o365/issues/6, there is a lot of new code that could be better tested.

Hopefully this is useful! Let me know if this there is anything you wish me to change (scope of tests, size of PR, code style, etc.) and I'll gladly change.

If these tests seem like a good idea, I recommend adding one of [https://github.com/getsentry/responses](responses) or [https://vcrpy.readthedocs.io/en/latest/](vcrpy) as a dev requirement, since they will provide more robust network mocking. The basic implementation in this PR is just to avoid adding a new dependency.

Coverage before

| Name            | Stmts | Miss | Cover |
| -               | -     | -    | -     |
| O365/message.py | 550   | 410  | 25%   |

Coverage after

| Name            | Stmts | Miss | Cover |
| -               | -     | -    | -     |
| O365/message.py | 550   | 121  | 78%   |

### DIFF
--- a/O365/utils/attachment.py
+++ b/O365/utils/attachment.py
@@ -468,11 +468,12 @@ class BaseAttachments(ApiComponent):
         #  select and then download one by one.
         return True
 
-    def _update_attachments_to_cloud(self, chunk_size=DEFAULT_UPLOAD_CHUNK_SIZE):
+    def _update_attachments_to_cloud(self, chunk_size=None):
         """ Push new, unsaved attachments to the cloud and remove removed
         attachments. This method should not be called for non draft messages.
         """
         # ! potentially several api requests can be made by this method.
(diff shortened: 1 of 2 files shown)


janscas:
Great work, this is very useful to test commits and don't break anything.
This project currently doesn't have any CD/CI pipelines so test are not being checked.

Unfortunately I don't have the time to do this, but any work done will very welcome here.

Thank a lot

strokirk:
Glad to be of use!

Issue #780, opened 2022-05-13
OneDrive download saves as wrong extension when converted to PDF


Andrew-Cha:
# Issue
When using the OneDrive download method and using the code
```
file.download(to_path = "/home/user/Desktop/", convert_to_pdf = True)
```
The expected behavior for a file is to convert "SampleFile.xlsx" from OneDrive to `"/home/user/Desktop/SampleFile.pdf"`
The actual outcome is ``"/home/user/Desktop/SampleFile.xlsx"`; Renaming the file to use the extension `.pdf` makes it open well.

When using the OneDrive download method and using the code
```
file.download(to_path = "/home/user/Desktop", name = "SampleFile.pdf", convert_to_pdf = True)
```

The expected behavior for a file is to convert "SampleFile.xlsx" from OneDrive to `"/home/user/Desktop/SampleFile.pdf"`
The actual outcome is a runtime error due to if statements looking for the extension in the *overwritten* local variable`name`, instead of `self.name` at line 87 of `drive.py`

# Source of Issue
Begins at line 52 and ends at line 87 of `drive.py`. Follow the code logic under the assumption that you will upload a mysterious file type (xlsx, ppt, docx) to OneDrive and will download it in two steps total (upload from path -> download as pdf to path)

# Workaround
```
file.download(name = file_path, output = filestream, convert_to_pdf = True)
```

Using the output parameter and passing the "name" as the original file path, or anything that includes the original file path fixes this. The convoluted flow of path checking can be seen all in `drive.py`. I am currently using this workaround, but would appreciate a fix to simplify the understanding of the library.

Andrew-Cha:
Thanks. @stanislau-arkhipenka

PR #797, opened 2022-07-04
bugfix for convert_to_pdf function

stanislau-arkhipenka:
This is a fix for Issue #780
Without this fix when file is downloaded with convert_to_pdf == True - it downloaded with original extension (like .xlsx)
With this fix if convert_to_pdf == True extension of a file will be replaced with .pdf


janscas:
Thanks!

Issue #804, opened 2022-07-16
How to access "inferenceClassification" in msg object


esiebomajeremiah:
So I am working on an application and I need to access the "inferenceClassification" value of a message. 
this value tells me whether a  message is focused or not.

There seem to be no way to access this value from this library even though it is provided by the MS graph API
it is also contained in the cloud_data within the code

but there is no way to access it on the message object

Below is a extract from the data returned by the graph API (cloud_data)
we can see that it contains the inferenceClassification property.
```
{
    .
    .
    .
    "inferenceClassification": "focused",
    "body": {
        "contentType": "html",
        "content": '<html><head>\r\n<meta http-equiv="Content-Type" content="text/html; charset=utf-8"></head><body><p>Hello Compose</p></body></html>',
    },
    "toRecipients": [
        {
            "emailAddress": {
                "name": "esiebomaj@gmail.com",
                "address": "esiebomaj@gmail.com",
            }
        }
    ],
    "ccRecipients": [],
    "bccRecipients": [],
    "replyTo": [],
    "flag": {"flagStatus": "notFlagged"},
}
```

So my issue is, is there an alternative way to access the property from the msg object 
or is there a way to get access to the cloud_data so that properties that are not taken care of by the library can be handled manually. 

Thanks!!!!

esiebomajeremiah:
@janscas pls help 🙏

BlueSideStrongSide:
are you online? I think I can see where you may be able to make your edit

BlueSideStrongSide:
https://github.com/O365/python-o365/blob/904b33f89e846f515ba131cd6449ff1115061870/O365/message.py#L258
^ I would start there and see if you can add another attribute to just export the entire object into the name space... for debug purposes to start. 

If that works you should be okay to submit a PR and just add a new name space with that property called out, using the same .get logic. 

I do not have this all setup for testing, but I think a few static prints and some inspection you should be able to sort through your issue. 

If I get some time to spin up a test... I may poke at it more. Trying to try and contribute to open source more this year, so forgive me if this message is a bit random discord'ish :)

janscas:
PR are welcome. Just add a property to message to return that from and internal attribute.  Set that internal attribute from cloud data.

BlueSideStrongSide:
Should have one shortly, doing a bit of research on the PR process :)

esiebomajeremiah:
@BlueSideStrongSide Thanks for working on this!

esiebomajeremiah:
[...]

PR #805, opened 2022-07-16
Add access to inferenceClassification in msg object

BlueSideStrongSide:
update for issue: How to access "inferenceClassification" in msg object #804

We add a new instance attrib and create a new property for this attrib to be retrieved.


janscas:
Great! Thanks

PR #820, opened 2022-08-13
Added ROPC authentication flow. Added environmental variable token st…

pierfrancesto:
…orage system. Added several objects to fully interact with a Microsoft Planner of a specific group

### DIFF
--- a/O365/__init__.py
+++ b/O365/__init__.py
@@ -8,7 +8,7 @@ from .__version__ import __version__
 
 from .account import Account
 from .connection import Connection, Protocol, MSGraphProtocol, MSOffice365Protocol
-from .utils import FileSystemTokenBackend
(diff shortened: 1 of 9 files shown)


pierfrancesto:
Hi @janscas! The purpose of this commit is to provide some objects to interact with Microsoft Planner. Since you need a delegated permission type to work with the planner API, I added a ROPC authentication flow which is convenient because is a one step authentication process (https://docs.microsoft.com/en-gb/azure/active-directory/develop/v2-oauth-ropc). Furthermore, I added an environmental variable token storage object, just to provide an alternative to the local file storage system. By reading the tests everything should be clear, if it is not the case please let me know. Hope you'll like the changes

janscas:
Hi this is great!

PR #830, opened 2022-09-06
Add support for "certificate" authentication

dieresys:
This method utilizes a JWT assertion to obtain the access token, as described in:
https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#second-case-access-token-request-with-a-certificate.

An example is also included with a helper to create the JWT assertion using the Private Key (PEM format), as described in: https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials

This helper uses pyjwt. Since we need to encode the assertion with RS256, the 'cryptography' module is also required, as described in: https://pyjwt.readthedocs.io/en/stable/usage.html#encoding-decoding-tokens-with-rs256-rsa

### DIFF
--- a/O365/account.py
+++ b/O365/account.py
@@ -34,12 +34,12 @@ class Account:
             # convert the provided scopes to protocol scopes:
             if scopes is not None:
                 kwargs['scopes'] = self.protocol.get_scopes_for(scopes)
-        elif auth_flow_type == 'credentials':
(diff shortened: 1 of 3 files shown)


janscas:
That's great! much appreciated

PR #836, opened 2022-09-15
bug fix on update method

pierfrancesto:
update fixed, before it was not able to update with an empty list or dict. plus some adjustments


janscas:
Thanks

Issue #838, opened 2022-09-19
problem with the query of getting all-day type events on a day


agn-7:
I have an issue with the query of getting all-day type events on a day. Actually, I'm going to fetch the events on a specific day by a query, and the problem is I get an extra all-day event from the next day as well!
My ms-calendar timezone is set to Istanbul; subsequently, the query's start date and end date are also in the Istanbul timezone.

Here's what I've done so far and the response:

```python
...
start_date = pytz.timezone("Europe/Istanbul").localize(datetime(2022, 9, 18))
end_date = pytz.timezone("Europe/Istanbul").localize(datetime(2022, 9, 19))
q = calendar.new_query('start').greater_equal(start_date)
q.chain('and').on_attribute('end').less(end_date)
events = calendar.get_events(query=q, include_recurring=True)
pprint(
    [
        (
            g.to_api_data()["subject"],
            g.to_api_data()["start"],
            g.to_api_data()["end"]
        ) for g in events])
```

Out:

```
[('dummy 2',
  {'dateTime': '2022-09-18T03:00:00', 'timeZone': 'Turkey Standard Time'},
  {'dateTime': '2022-09-19T03:00:00', 'timeZone': 'Turkey Standard Time'}),
 ('event 4',
  {'dateTime': '2022-09-18T23:30:00', 'timeZone': 'Turkey Standard Time'},
  {'dateTime': '2022-09-19T00:00:00', 'timeZone': 'Turkey Standard Time'}),
 ('event 2',
  {'dateTime': '2022-09-19T00:00:00', 'timeZone': 'Turkey Standard Time'},
  {'dateTime': '2022-09-19T00:30:00', 'timeZone': 'Turkey Standard Time'}),
 ('dummy 3',
  {'dateTime': '2022-09-19T03:00:00', 'timeZone': 'Turkey Standard Time'},
  {'dateTime': '2022-09-20T03:00:00', 'timeZone': 'Turkey Standard Time'})]
```

As you can see in the screenshot my expectation would be getting `dummy 2`, `event 2`, and `event 4`, but I'm getting `dummy 3` as well which is into the next day!

![2022-09-19_12-01](https://user-images.githubusercontent.com/14202344/190984614-e7391c84-cc85-43d1-81bb-1580ebbc5161.png)

agn-7:
Hi @janscas

Could you please look into this issue?

agn-7:
I bypassed this issue as follows — adding 1 second forward and backward:

```python
[...]

PR #843, opened 2022-09-29
fixed the all-day events time representation

agn-7:
According to this [issue](https://github.com/O365/python-o365/issues/838) (#838 ), the representation datetime of an all-day event was not correct as +3 house was added to an aware-datetime with the turkey timezone. This PR will resolve this issue.


janscas:
Thanks!

Issue #795, opened 2022-06-20
Unable to do basic Excel operations


Kornosky:
Hello! I haven't seen anyone else run into this issue, but no matter what I do I can't get past this error when trying to do simple actions like "get_row". I'm stumped on how to fix it.
Main error:
```Error Message: The API you are trying to use could not be found. It may be available in a newer version of Excel. ```
Full Error:
```
Traceback (most recent call last):
  File "c:/Users/chris/Desktop/DeleteMe/TestCSVTeams/from O365.excel import WorkBook-bak.py", line 130, in <module>
    main()
  File "c:/Users/chris/Desktop/DeleteMe/TestCSVTeams/from O365.excel import WorkBook-bak.py", line 103, in main
    print(table.get_row_at_index(0))
  File "C:\Users\chris\AppData\Local\Programs\Python\Python36-32\lib\site-packages\O365\excel.py", line 1337, in get_row_at_index
    response = self.session.post(url, data={'index': index})
  File "C:\Users\chris\AppData\Local\Programs\Python\Python36-32\lib\site-packages\O365\excel.py", line 142, in post
    return self.con.post(*args, **kwargs)
  File "C:\Users\chris\AppData\Local\Programs\Python\Python36-32\lib\site-packages\O365\connection.py", line 822, in post
    return self.oauth_request(url, 'post', data=data, **kwargs)
  File "C:\Users\chris\AppData\Local\Programs\Python\Python36-32\lib\site-packages\O365\connection.py", line 800, in oauth_request
    return self._internal_request(self.session, url, method, **kwargs)
  File "C:\Users\chris\AppData\Local\Programs\Python\Python36-32\lib\site-packages\O365\connection.py", line 762, in _internal_request
    raise HTTPError('{} | Error Message: {}'.format(e.args[0], error_message), response=response) from None
[...]

PR #849, opened 2022-10-02
Update excel.py

jeroenvermunt:
Fixes issue mentioned in #795 

source: https://stackoverflow.com/questions/65210929/ms-graph-api-update-tablerow-fail-for-400-client-error-bad-request-for-url

### DIFF
--- a/O365/excel.py
+++ b/O365/excel.py
@@ -966,7 +966,7 @@ class TableRow(ApiComponent):
             getattr(parent, 'main_resource', None) if parent else None)
 
         # append the encoded column path
-        main_resource = '{}/rows/{}'.format(main_resource, self.object_id)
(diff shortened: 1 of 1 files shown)


ShadiNM:
Hello @alejcas I think the same fix should be applied in get_row_at_index. It still uses the POST with index in the payload instead of GET with the index passed like /rows/itemAt(index={index})

Will you have time to add this?

def get_row_at_index(self, index):
        """
        Returns a table row by it's index
        :param int index: the zero-indexed position of the row in the table
        """
        if index is None:
            return None

        url = self.build_url(self._endpoints.get('get_row_index'))
        response = self.session.post(url, data={'index': index})

        if not response:
            return None

PR #858, opened 2022-10-13
add Pagination to Plan.list_tasks

hcallen:
Adds Pagination return to Plan's list_tasks method to allow the return of all tasks instead of only 400 (endpoint limit). Also, it appears the endpoint doesn't support the "$top" query parameter.


janscas:
great thanks!

PR #870, opened 2022-11-05
Added ability to set/ disable auto reply

lodesmets:
These changes allow to set/ disable the auto reply.
With this you can schedule the auto reply functionality in outlook

### DIFF
--- a/O365/mailbox.py
+++ b/O365/mailbox.py
@@ -516,9 +516,58 @@ class Folder(ApiComponent):
 
 class MailBox(Folder):
     folder_constructor = Folder
+    _endpoints = {
+        'settings': '/mailboxSettings',
+    }
 
     def __init__(self, *, parent=None, con=None, **kwargs):
         super().__init__(parent=parent, con=con, root=True, **kwargs)
+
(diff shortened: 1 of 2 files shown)


janscas:
Great addition! thanks!

RogerSelwyn:
@lodesmets & @janscas This PR broke many of the Mailbox endpoints by overriding the Folder endpoints array. Consequently a call to Mailbox.get_folders() no longer works, along with a number of others. `_endpoints` in `Mailbox` needs to be added to those from `Folder`.

lodesmets:
> @lodesmets & @janscas This PR broke many of the Mailbox endpoints by overriding the Folder endpoints array. Consequently a call to Mailbox.get_folders() no longer works, along with a number of others. `_endpoints` in `Mailbox` needs to be added to those from `Folder`.

I'm terribly sorry for breaking the feature. I didn't notice that mailbox (unlike other classes wasn't a child of ApiComponent).
But I noticed you already added a PR for the fix

janscas:
@RogerSelwyn true! I didn’t even notice… i’m sorry

@lodesmets don’t worry this happens…

PR #871, opened 2022-11-08
Add headers to emails before sending

ebeers-png:
I'd like to be able to send mail with custom headers. Currently, Messages can be initialized with header data through cloud data, but the field `message_headers` isn't included in the sending process. I made a simple edit to include `message_headers` in the data that gets sent with the post request. 

To add a header to a message before sending:
```python
message.message_headers = [{
    'name': 'X-customheader',
    'value': 'headervalue'
}]
```
The format of the dict is specified on [this page](https://learn.microsoft.com/en-us/graph/api/resources/internetmessageheader?view=graph-rest-1.0) of Microsoft's documentation, and the name must start with "X-" or "x-". 

Let me know if there are other considerations to be mindful of here that I'm missing. Thanks!

### DIFF
--- a/O365/message.py
+++ b/O365/message.py
@@ -681,6 +681,9 @@ class Message(ApiComponent, AttachableMixin, HandleRecipientsMixin):
                 if key not in restrict_keys:
                     del message[key]
 
+        if self.message_headers:
+            message[cc('internetMessageHeaders')] = self.message_headers
+
         return message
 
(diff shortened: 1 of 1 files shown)


janscas:
I would like to check that this does not break a message when it's a draft that is to be sended...

Did you tried that?

Also, message_headers should respect tracking changes?
I mean you are including them always no matter if the message_headers have changed or not

PR #874, opened 2022-11-12
Added support for tasks/todo via MS Graph API

RogerSelwyn:
As promised here is a MS Graph api version of the Tasks/ToDo.

At this time, I've:
- implemented as a new class file `tasks_graph.py`, primarily because my formatting style on VSCode will re-format the rest of the code in `tasks.py`. I can copy the code into the original class file and rename the classes to names such as `ToDoGraph` if you wish.
- renamed a couple of attributes. (e.g. in the Graph api it uses `title` for the task, rather than `subject`). Also some constants are now all lower case (e.g. `Completed` has become `completed`).
- endpoints to constants - I'm paranoid about not matching things up. I can change back if preferred.
- tried to exercise all routes through the code. So hopefully no problems
- implemented a filter capability on `get_tasks` not present in existing code. This allows you to get uncompleted tasks for instance.

I think in passing I have fixed some issues that existing in MS Office implementation. Certainly there was at least one redundant endpoint, but I think there was at least one utilised endpoint that didn't have the relevant parameters passed to it. I'd have to go back and try and test using Office API to validate that statement. 

There seems to be no existing documentation for Tasks. I can potentially add something to the Readme if you like. Or have a looks at how the generated docs work to see if it will generate meaningfully.

### DIFF
--- a/O365/account.py
+++ b/O365/account.py
@@ -255,13 +255,12 @@ class Account:
     def tasks(self, *, resource=''):
         """ Get an instance to read information from Microsoft ToDo """
 
-        if not isinstance(self.protocol, MSOffice365Protocol):
-            raise RuntimeError(
-                'todo api only works on Microsoft Office 365 API')
+        if isinstance(self.protocol, MSOffice365Protocol):
+            from .tasks import ToDo
+        else:
+            from .tasks_graph import ToDo as ToDo
 
-        from .tasks import ToDo
         return ToDo(parent=self, main_resource=resource)
-
     
     def teams(self, *, resource=''):
         """ Get an instance to read information from Microsoft Teams """
--- a/O365/tasks_graph.py
+++ b/O365/tasks_graph.py
@@ -1,0 +1,733 @@
+"""Methods for accessing MS Tasks/Todos vis the MS Graph api."""
+import datetime as dt
+import logging
+
+# noinspection PyPep8Naming
+from bs4 import BeautifulSoup as bs
+from dateutil.parser import parse
+
+from .utils import ApiComponent, TrackerSet
+
+log = logging.getLogger(__name__)
+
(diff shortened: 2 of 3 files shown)


RogerSelwyn:
I should add whilst the code is very similar to the Office APIs version, there are variations throughout. Some outlined above, but also items such as how the Graph Api identifies the default folder (for which there is no longer a straight API call, you have to retrieve all and enumerate - I can probably improve that ny using a filter on the relevant attribute).

janscas:
Hi, thanks for this PR.

Great work and thanks for updating the readme!

As for your comments:

- I don't mind having this in a separate file. So it's ok as it is now.
- About the different attribute names between Graph and Office protocols I would like to unite them but It's ok for the moment as it is.
- About the other api differences, O365 should try to do the best for the user and seems it's already doing that. Maybe you could catch the default folder but it won't probably make any difference in many use cases.
- Using constants for the endpoints keys is a great addition,

janscas:
I'm going to release a new version now with these changes

RogerSelwyn:
If you wait to tomorrow I can change back the different attribute names. I thought about doing it this morning.

janscas:
Ok great!
I'll commit a few pep8 fixes to your code

RogerSelwyn:
Ok, I don’t use pep8. Home Assistant uses flake8 and black. 

The code supports default_folder. I need to check what I’ve done with Status, probably add an enumerator or something to keep it the same as O365.

janscas:
a few fixes here: 9702c3b225ffcd3ef712fddffe27d1908b847012

RogerSelwyn:
I’ll need to do a new PR I think?

janscas:
> I’ll need to do a new PR I think?

yep

PR #881, opened 2022-11-18
:sparkles: Support for immutable id's on messages (mailbox)

NielsDebrier:
I would like to receive an immutable id when retrieving a message (mailbox).
The id's for a message in O365 changes every time they get moved to a different folder. This makes it very hard to keep track of a message throughout the mailbox. We have had issues with this in the past and recently have once again the need for an immutable id.

This PR should provide an easy abstraction and support this use-case.
[Microsoft graph api documentation](https://learn.microsoft.com/en-us/graph/outlook-immutable-id)

If there are any questions or you like to see some changes, please let me know.

Original commit message: 
> Changes so users can request messages with an immutable id that will never change. If you are using the mailbox integration having to deal with ever changing id's is very difficult. With this change users can just set a param and get an immutable id from o365.

### DIFF
--- a/O365/mailbox.py
+++ b/O365/mailbox.py
@@ -136,7 +136,8 @@ class Folder(ApiComponent):
         else:
             return folders
 
-    def get_message(self, object_id=None, query=None, *, download_attachments=False):
+    def get_message(self, object_id=None, query=None, *, download_attachments=False,
+                    immutable_id=False):
         """ Get one message from the query result.
          A shortcut to get_messages with limit=1
         :param object_id: the message id to be retrieved.
@@ -144,6 +145,7 @@ class Folder(ApiComponent):
          "displayName eq 'HelloFolder'"
         :type query: Query or str
         :param bool download_attachments: whether or not to download attachments
+        :param bool immutable_id: whether or not to get an immutable id
         :return: one Message
         :rtype: Message or None
         """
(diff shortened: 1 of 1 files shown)


janscas:
Hi, thanks for this PR.

I think this will lead you to some problems. You are only requesting inmutable Ids from get messages methods from folders, but message id's are used in message.py as well.

For example `message.send` will set a `message.object_id` that is not inmutable. Also message.move, etc.

I know you may not need this, but I think it's pretty needed to provide a method to get consistent inmutable Ids.

Also I notice this header can be set to get also inmutable ids for contacts and etc.

I think there should be a better way to handle this.

Maybe in `Connection` init allow to pass `default_headers` param that will be stored as an attribute in connection then change line [720/721](https://github.com/O365/python-o365/blob/9a7c07e160f77ab3e0feaec05fea659dc5814597/O365/connection.py#L720) to this:

```python
if 'headers' not in kwargs:
    kwargs['headers'] = {**self.default_headers}
else:
    for key, value in self.default_headers.items():
        if key not in kwargs['headers`]:
            kwargs['headers'][key] = value
```

By doing this your good to go and also provides with a common method to set other default headers.

```python
account =Account(..., default_headers={'Prefer': 'IdType="ImmutableId"'})
``` 


This is **not tested** and may have other implications. 
**I don't know if such a header will be accepted on all api calls.**

Can you take a look and provide feedback?

Thanks

NielsDebrier:
Thank you for your feedback.

When I have the time I will try to look into it and make the changes.

yeyeric:
hello,
just wondering if there was a solution to this as I need immutable ID for my use cases as well ?

yeyeric:
based on your post @janscas  which was almost complete, I've made a PR: https://github.com/O365/python-o365/pull/927
Tested here and working as expected, adding the default_header while creating the account makes microsoft to work with immutable ID (strange it's not the default behaviour though).

PR #920, opened 2022-12-21
Extend mailbox settings

RogerSelwyn:
This PR extends the setting of auto_reply status, to setup for for full mailbox settings retrieval.

- Enables getting of autoreply settings as a subclass to mailbox settings (as it is in the MS Graph API - https://learn.microsoft.com/en-us/graph/api/user-get-mailboxsettings?view=graph-rest-1.0&tabs=http)
- Once retrieved, the settings can then be altered individually.
- Alters mailbox.set_automatic_reply/set_disable_reply to use the new classes, but does not alter the external calls to those methods

This should be easily extensible to retrieve all mailbox settings.

**Query** I've set the automaticrepliessettings class to be an ApiComponent. It has no direct api calls through it, they are all handled at MailboxSettings, so not sure if it needs to be an ApiComponent, but had to set it that way to enable cloud_data_key.

**Note** I submitted a baseline commit, because my setup, I think, formats with a slightly longer line length. So this will give you greater clarity on what I've actually changed if you look at the second commit.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -35,6 +35,7 @@ DEFAULT_SCOPES = {
     'basic': [('offline_access',), 'User.Read'],
     'mailbox': ['Mail.Read'],
     'mailbox_shared': ['Mail.Read.Shared'],
+    "mailbox_settings": ["MailboxSettings.ReadWrite"],
     'message_send': ['Mail.Send'],
     'message_send_shared': ['Mail.Send.Shared'],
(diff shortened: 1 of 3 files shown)


janscas:
I was on vacation! yep I'll do

RogerSelwyn:
> I was on vacation! yep I'll do

Excellent!

janscas:
released

RogerSelwyn:
Thanks!

PR #927, opened 2023-01-20
Update connection.py to allow default_headers to allow immutable ID

yeyeric:
Allow to set default_headers when creating account so that those params are sent.
Typical usage is default_headers={"Prefer": 'IdType="ImmutableId"'} which allows to have permanent immutable id instead of default temporary ID which changes as soon as the object (message / event / task / contact) gets moved / updated.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -329,7 +329,9 @@ class Connection:
                  auth_flow_type='authorization',
                  username=None, password=None,
                  timeout=None, json_encoder=None,
-                 verify_ssl=True, **kwargs):
+                 verify_ssl=True,
+                 default_headers: dict = None,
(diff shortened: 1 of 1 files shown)


yeyeric:
hi @janscas , do you think this can be merged or needs further review ? this update is important for my project so that'd be great
Thanks

yeyeric:
hello @janscas , can this be pushed through or any specific procedure to follow  (first PR for me on a public repo) ?
thanks

janscas:
Hi, first, thanks for this PR.

For me it's ok, but I don't have the time to test this.

Hope you did!

janscas:
2.0.26 released

Issue #937, opened 2023-02-19
[Bug] can not update Table show_headers and show_totals of Table


bfcs:
in excel.py -> Table -> update(), if show_headers or show_totals is False, then it won't be added into request, suppose to change to 
```python
  if show_headers is not None:
      ...
```

<img width="793" alt="image" src="https://user-images.githubusercontent.com/52602045/219908136-206f5983-6f57-4a4f-91f4-b7d166f0cb59.png">

janscas:
Can you make a PR fixing it? Thanks

bfcs:
I am glad to

janscas:
Thanks!

PR #938, opened 2023-02-20
fix the bug can not update Table show_headers and show_totals of Table

bfcs:
in excel.py -> Table -> update(), if show_headers or show_totals is False, then it won't be added into request, mentioned in [Issue 937](https://github.com/O365/python-o365/issues/937)


janscas:
Thanks!!

PR #941, opened 2023-03-06
Rename UA Kyiv timezone

tooSadman:
Renamed "Europe/Kiev" to "Europe/Kyiv" as per ICANN's change: https://mm.icann.org/pipermail/tz-announce/2022-August/000071.html

Fixes error when using calendar with timezone "Europe/Kyiv":
`pytz.exceptions.UnknownTimeZoneError: "Can't find Iana TimeZone Europe/Kyiv"`


janscas:
Thanks!

PR #952, opened 2023-04-04
Add support for filter by due datetime in Tasks

RogerSelwyn:
Added extra mapping in Utils to support filter by due datetime and reminder datetime in Tasks

Also added extra detail to connection error reporting <-- This can be removed if preferred but it provides extra information on why a call failed.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -792,14 +792,22 @@ class Connection:
                 try:
                     error = response.json()
                     error_message = error.get('error', {}).get('message', '')
+                    error_code = (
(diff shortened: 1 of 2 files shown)


RogerSelwyn:
Do you know when you will release a new version with this included?

janscas:
I can do it in a few days

RogerSelwyn:
> I can do it in a few days

Thanks

janscas:
done

RogerSelwyn:
> done

Thanks

PR #956, opened 2023-04-22
Adding option to set file created and last modified time while uploading

yeyeric:
Hi @janscas ,

I've just added an option to set file created datetime and last modified datetime while uploading the file.

This has been tested and all working perfectly with all cases (with conflict handling, without, with only created date time, or last modified datetime)

Thank you


janscas:
Great addition.

Thanks!

PR #960, opened 2023-04-30
add missing format keyname for invoke_function endpoint

juanboro:
Hi - it seems that the format key was missing for building the url for invoke_function.  This makes it work for me.


janscas:
Thanks

PR #974, opened 2023-06-08
Pass keyword arguments to request_token also when auth_flow_type is credentials, certificate or password

IAL32:
When authenticating with "credentials", "certificate" or "password", there is no way to control the behaviour of `request_token` by passing extra keyword arguments. This is already possible for "authorization" and "public" authentication types.
This PR fixes that by passing the keyword arguments also when auth_flow_type in "credentials", "certificate" or "password".


janscas:
Thanks!

PR #994, opened 2023-08-02
Removed dependency from pytz

ponquersohn:
Created the initial version without pytz.
Tested internally and the exceptions are gone.


ponquersohn:
@janscas We have tested it internally on multiple scenarios. No errors so far. Probably ready to be merged.

O365/utils/utils.py
```diff
@@ -947,8 +948,7 @@ def _parse_filter_word(self, word):
             if isinstance(word, dt.datetime):
                 if word.tzinfo is None:
                     # if it's a naive datetime, localize the datetime.
-                    word = self.protocol.timezone.localize(
-                        word)  # localize datetime into local tz
+                    word = word.replace(tzinfo=self.protocol.timezone)  # localize datetime into local tz
                 if word.tzinfo != pytz.utc:
```
janscas, line 952: This still mentions pytz

O365/utils/utils.py
```diff
@@ -947,8 +948,7 @@ def _parse_filter_word(self, word):
...
                     word = word.astimezone(
                         pytz.utc)  # transform local datetime to utc
```
janscas, line 954: Here also

O365/utils/utils.py
ponquersohn: Yea, not sure how i missed it. Fixed.

ponquersohn:
@janscas Im unable to test on Windows. Sorry. 
My internal apps are using the code from PR since the PR was open without issues.

janscas:
I'll merge this as it is

Hope for no bugs

misilot:
@janscas thank you for merging this! Any chance we could get a new tagged version or release? Thanks again!

janscas:
Done 2.0.29

misilot:
Thanks. 

I am still seeing the following in 2.0.31

>.venv/lib/python3.10/site-packages/O365/utils/windows_tz.py:638: PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html on Linux

misilot:
So it looks like it's from tzlocal, and needs to be upgraded to 5.0. 

From https://pypi.org/project/tzlocal/: 

> 
> tzlocal 4.0 also adds an official function get_localzone_name() to get only the timezone name, instead of a timezone object. On unix, it can raise an error if you don’t have a timezone name configured, where get_localzone() will succeed, so only use that if you need the timezone name.
> 
> 4.0 also adds way more information on what is going wrong in your configuration when the configuration files are unclear or contradictory.
> 
> Version 5.0 removes the pytz_deprecation_shim, and now only returns zoneinfo objects, like verion 3.0 did. If you need pytz objects, you have to stay on version 4.0. If there are bugs in version 4.0, I will rekease updates, but there will be no further functional changes on the 4.x branch.

PR #996, opened 2023-08-16
fix key missing innerError when 404

yeyeric:
.get is sending an empty string instead of empty dict so that next get is working


yeyeric:
i was using older version, has already been fixed

PR #999, opened 2023-08-31
Add Support for Disabling Forward Button in Calendar Invitations

Gregorek85:
Hi,

I have prepared a small modification that allows users to disable the forward button in calendar invitations for a given event. This feature can be useful in scenarios where the event organizer wants to restrict the distribution of the event details.

Changes Made:
Added a new attribute __no_forwarding initialized to False.
Included SingleValueExtendedProperties in the to_api_data method to handle the "Do Not Forward" setting.
Added a property no_forwarding with its corresponding setter method to enable or disable this feature.
Usage:
To disable the forward button for an event, simply set the no_forwarding attribute to True:
```python
event.no_forwarding = True
```

I am attaching a screenshot to show how this looks in the event invitation:
Screenshot


janscas:
thanks!

PR #1013, opened 2023-10-09
Fix typos causing home address to be unretrievable.

### LINKED ISSUE #989: Bug: Unable to retrieve or add homeAddress from Contacts(address_book.py)
Hi, 

In the current state it is not possible to retrieve or add homeAddress, because of a simple typo.

address_book.py line 94: self.__home_address = cloud_data.get(cc('homesAddress'), {})
should be:
self.__home_address = cloud_data.get(cc('homeAddress'), {})

as well as on line 372: self._track_changes.add(self._cc('homesAddress'))
should be:
self._track_changes.add(self._cc('homeAddress'))

and the last one on line 473: cc('homesAddress'): self.__home_address,
should be:
cc('homeAddress'): self.__home_address,

kwilsonmg:
This is a rather simple change that would fix issue #989.

### DIFF
--- a/O365/address_book.py
+++ b/O365/address_book.py
@@ -91,7 +91,7 @@ class Contact(ApiComponent, AttachableMixin):
             # a Contact from OneDrive?
             self.__emails.add(email)
         self.__business_address = cloud_data.get(cc('businessAddress'), {})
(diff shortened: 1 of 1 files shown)


janscas:
Thanks!

PR #1017, opened 2023-10-10
Update drive.py for the group_id

Vaishnavi21721:


janscas:
Can you make granted_to hold a tuple (display_name, id)?

Issue #1025, opened 2023-11-18
Reading content of .msg and .eml type attachments is not possible


kamalvk18:
Currently what I can see is we have a method defined for saving email type attachments by using 
`message.attachments.save_as_eml(attachment)`
But there is no method defined for just reading the content. In my case I want to read the content of the .eml type attachment and store it as bytes for further processing.

Please let me know if there is already a possible way to read those attachments.

kamalvk18:
I've raised a PR to resolve this. Can you please review and add comments if you find anything to be modified. This is really important for our project.

alejcas:
Merged ;)

PR #1026, opened 2023-11-19
To solve issue #1025 (Being able to read mime content of .eml or .msg type attachments)

kamalvk18:
This push is to add a feature where people can read the contents of the email type attachments.

Current code base supports only saving of the email type attachments, but there may be cases where people might want to read the content of the email type attachments for further processing. This push exactly adds that feature.


alejcas:
Thanks!

Issue #872, opened 2022-11-09
Check_status of copied DriveItem not working as intended


jeroenvermunt:
I want to modify a copied DriveItem like so:

````
copy = calculators.copy(folder)

for status, perc_complete in copy.check_status():
    print(status, perc_complete)

item = copy.get_item()
````

However, the check_status() method seems to fail on me, returning the following:

````
Client Error: 401 Client Error: Unauthorized for url: https://my_domain.sharepoint.com/_api/v2.0/drives/{drive_id}/items/item_id}) | Error Message: Invalid audience Uri 'https://graph.microsoft.com/'.
````

This Error arises as a requests.exceptions.HTTPError when running `CopyOperation._request_status()`. I highlighted the line below. It seems as the `monitor_url` breaks when the copy is complete, whilst not properly finishing the `_request_status()` method. I tried catching the error around the relevant line, but can't find a good way to get the driveItem in the form of `https://graph.microsoft.com/beta/drive/{drive_id}/items/{item_id}`. Any ideas to solve this issue?

````
def _request_status(self):
        """ Checks the api endpoint to check if the async job progress """

        if self.item_id:
            return True

        response = self.con.get(self.monitor_url)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            
        if not response:
            return False

        data = response.json()

        self.status = data.get('status', 'inProgress')
        self.completion_percentage = data.get(self._cc('percentageComplete'),
                                              0)
            
        self.item_id = data.get(self._cc('resourceId'), None)

        return self.item_id is not None
````

janscas:
I don't have a clue right now... sorry

will-byrne-cardano:
Hi @alejcas @jeroenvermunt , 

I encountered this issue yesterday and after some debugging I found the following issues:

[...]

PR #1033, opened 2023-12-21
Fix bug with `get_item` on `CopyOperation`

will-byrne-cardano:
method `CopyOperation.get_item` only worked if the copied item was copied to the same location as the original item. Here we introduce the optional argument target which defaults to the parent if not given (i.e. the scenario where we copy to the same location as original item) so that the copied item can be successfully retrieved from it's new location upon successful copy completion. 

https://github.com/O365/python-o365/issues/872 

Also, update mistake in comment on parent type in the init

### DIFF
--- a/O365/drive.py
+++ b/O365/drive.py
@@ -128,12 +128,14 @@ class CopyOperation(ApiComponent):
         'item': '/items/{id}',
     }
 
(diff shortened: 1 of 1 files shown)


alejcas:
merged! thanks again

Issue #1011, opened 2023-10-05
PYTZ removal causes errors in latest release


RogerSelwyn:
I'm trying to upgrade to the latest O365 release (2.0.31), with no changes to my code I get this error:

`TypeError: timezone() argument 1 must be datetime.timedelta, not str`

The call I'm making is this:

`account = Account( credentials, token_backend=token_backend,  timezone=("UTC"), main_resource=main_resource )`

If I remove the timezone from the call:

`account = Account( credentials, token_backend=token_backend, main_resource=main_resource )`

I get this error:

```
2023-10-05 17:33:42.688 WARNING (MainThread) [py.warnings] /HADev/Hassio/dev-config/O365/utils/windows_tz.py:638: PytzUsageWarning: The zone attribute is specific to pytz's interface; please migrate to a new time zone provider. For more details on how to do so, see https://pytz-deprecation-shim.readthedocs.io/en/latest/migration.html
  iana_tz.zone if isinstance(iana_tz, tzinfo) else iana_tz)
```

Is this expected, or is there a timezone parameter I can pass in which will remove the warning? I've tried passing in timedelta(microseconds=0) as implied in the first error, but I get the same as the second error.

It's probably me just missing something, but I can't see what it is.

RogerSelwyn:
If it helps, I'm using Python 3.11 on Debian Bookworm.

RogerSelwyn:
Just trying to debug this more. If I pass in UTC as the timezone (which is what get_windows_tz in windows_tz.py) is looking for, then line 100 (or they're about) in connection.py fails. In fact, as setup that line will always fail if it is reached. The If stamens before checks for a string, and if the string is passed to dt.timezone() it then errors.

```
        if timezone and isinstance(timezone, str):
            timezone = dt.timezone(timezone)
        try:
            self.timezone = timezone or get_localzone() 
```

RogerSelwyn:
The `get_localzone()`is returning a `_PytzShimTimezone` object that has a zone attribute, which when accessed triggers the warning in windows_tz.py.

RogerSelwyn:
Sorry for many posts, just doing my thoughts. Either it will error in connection.py if you pass in a string, or it will get a warning in windows_tz.py because it accesses the zone attribute, or it will error in windows_tz.py because the passed in timezone has no zone attribute. I don't see anyway to avoid at least a warning.

RogerSelwyn:
> I merged a few days ago a PR that dropped pytz from this project.
[...]

PR #1038, opened 2024-01-08
Add ability to set user status and get another users status

RogerSelwyn:
Add ability to set user presence status per application and set user preferred presence status, plus get another users presence status

Get my presence status is currently under Teams, so I have placed these there as well

### DIFF
--- a/O365/teams.py
+++ b/O365/teams.py
@@ -1,4 +1,5 @@
 import logging
+from enum import Enum
 
 from dateutil.parser import parse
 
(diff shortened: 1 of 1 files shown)


RogerSelwyn:
Blimey that was fast!! Do you know when you will release?

alejcas:
expect a release this week

RogerSelwyn:
Cool. looked like you were getting ready.

RogerSelwyn:
Just a comment, currently I can't move to a latest release because of this error - https://github.com/O365/python-o365/issues/1011


Unfortunately I only just looked at it again after a couple of months. You noted the issue as fixed, but is seems I still can't pass in a timezone as a string.

RogerSelwyn:
> Released 2.0.32

Thanks. Have you noted my problem with doing anything with setting dates? https://github.com/O365/python-o365/issues/1011

alejcas:
> > Released 2.0.32
> 
> Thanks. Have you noted my problem with doing anything with setting dates? #1011

Yes, but no clue for the moment

RogerSelwyn:
> Yes, but no clue for the moment

Are you able to replicate? I don't know whether it is just me... I've put some sample code on the issue. If you can replicate, then I know it's not just me being stupid :-(.

PR #1039, opened 2024-01-09
Fix get recurring instances events invalid start/end time bug

Invincibear:
Converts date objects into datetime objects in Event.get_occurrences() to add missing time parameters

```
  File "/var/www/domain.tld/backend/.venv/lib/python3.11/site-packages/O365/calendar.py", line 1378, in get_occurrences
    if start.tzinfo is None:
       ^^^^^^^^^^^^
AttributeError: 'datetime.date' object has no attribute 'tzinfo'
```


alejcas:
Thanks!

PR #1043, opened 2024-01-11
feat new feacture for timezone

Elq-eng:
I had a problem, because when the library responded to a message, it did not respond with the time zone of the area, for example Bogotá Colombia, so I wanted to add a new feature.

### DIFF
--- a/O365/message.py
+++ b/O365/message.py
@@ -1,6 +1,12 @@
 import datetime as dt
 import logging
 from enum import Enum
+from datetime import datetime, timedelta
+import logging
+from enum import Enum
+import re
+
+import pytz
 
 # noinspection PyPep8Naming
 from bs4 import BeautifulSoup as bs
@@ -755,6 +761,24 @@ class Message(ApiComponent, AttachableMixin, HandleRecipientsMixin):
             return None
 
         message = response.json()
+
+        # adding new feacture to the timezone 
+        html_content = message['body']['content']
+
+        pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
+
+        def change_sent_time(match):
+            sent_time_str = match.group(1)
+            sent_time = datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
+            # asign the time zone for example bogota
+            timezone_tz = pytz.timezone('America/Bogota')
+            sent_time_new = sent_time.replace(tzinfo=pytz.utc).astimezone(timezone_tz)
+            # return the new value
+            return f'<b>Sent:</b> {sent_time_new.strftime("%A, %B %d, %Y %I:%M:%S %p")}<br>'
+
+        data_modified = pattern.sub(change_sent_time, html_content)
+        # the new value is concatenated 
+        message['body']['content'] = data_modified
 
         # Everything received from cloud must be passed as self._cloud_data_key
         return self.__class__(parent=self, **{self._cloud_data_key: message})


O365/message.py
```diff
@@ -756,6 +762,24 @@ def reply(self, to_all=True):
...
+        # adding new feacture to the timezone 
+        html_content = message['body']['content']
+
+        pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
+
+        def change_sent_time(match):
+            sent_time_str = match.group(1)
+            sent_time = datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
+            # asign the time zone for example bogota
+            timezone_tz = pytz.timezone('America/Bogota')
```
Invincibear, line 774: We shouldn't be hard-coding a timezone, this needs to be a variable
Also I'm pretty sure `pytz` has been dropped due to `zoneinfo` requirements: https://github.com/O365/python-o365/blob/master/CHANGES.md#version-2030-2023-09-27

O365/message.py
```diff
@@ -756,6 +762,24 @@ def reply(self, to_all=True):
 
         message = response.json()
 
+        # adding new feacture to the timezone
```
Invincibear, line 765: This comment should be deleted, we can see the changes in the diff

O365/message.py
```diff
@@ -756,6 +762,24 @@ def reply(self, to_all=True):
...
+        html_content = message['body']['content']
+
+        pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
+
+        def change_sent_time(match):
+            sent_time_str = match.group(1)
+            sent_time = datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
+            # asign the time zone for example bogota
+            timezone_tz = pytz.timezone('America/Bogota')
+            sent_time_new = sent_time.replace(tzinfo=pytz.utc).astimezone(timezone_tz)
```
Invincibear, line 775: Another `pytz` reference

O365/message.py
```diff
@@ -756,6 +762,24 @@ def reply(self, to_all=True):
...
+            # return the new value
+            return f'<b>Sent:</b> {sent_time_new.strftime("%A, %B %d, %Y %I:%M:%S %p")}<br>'
```
Invincibear, line 777: To maintain support of older Python versions this library doesn't use `f-strings`, rewrite this using `.format()` instead

O365/message.py
```diff
@@ -1,6 +1,12 @@
 import datetime as dt
 import logging
 from enum import Enum
+from datetime import datetime, timedelta
+import logging
+from enum import Enum
+import re
+
+import pytz
```
Invincibear, line 9: O365 dropped `pytz`, see my comment below

O365/message.py
```diff
@@ -1,6 +1,12 @@
 import datetime as dt
 import logging
 from enum import Enum
+from datetime import datetime, timedelta
+import logging
+from enum import Enum
```
Invincibear, line 6: This is already imported on line 3

O365/message.py
```diff
@@ -1,6 +1,12 @@
 import datetime as dt
 import logging
 from enum import Enum
+from datetime import datetime, timedelta
+import logging
```
Invincibear, line 5: This is already imported on line 2

O365/message.py
```diff
@@ -756,6 +762,24 @@ def reply(self, to_all=True):
 
         message = response.json()
 
+        # adding new feacture to the timezone 
+        html_content = message['body']['content']
```
Invincibear, line 766: Variable name `html_content` could be more descriptive, something like `message_body_content_html` makes it instantly obvious what it contains and builds upon previously-defined `message` variable naming.

Invincibear (changes requested):
From what I understand, is the goal of this PR is to add the ability to modify the timezone of the sent property? Can you give an example of when this would need to be used?

Elq-eng:
> From what I understand, is the goal of this PR is to add the ability to inject (or modify) the timezone of the sent datetime?


More or less, the scenario I had was, when I tried to respond to a message with message.reply() it made the request but when I responded to it in my time zone it was 3:00 PM and in the email response it was 8 :00 PM

O365/message.py
```diff
@@ -755,7 +756,20 @@ def reply(self, to_all=True):
...
 
         message = response.json()
+        message_body_content_html  = message['body']['content']
 
+        pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
+
+        def change_sent_time(match):
+            sent_time_str = match.group(1)
+            sent_time = dt.datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
+            sent_time_new = sent_time.replace(tzinfo=dt.timezone.utc).astimezone(self.protocol.timezone)
```
Invincibear, line 766: We should be more descriptive with this variable name, lets use `sent_time_as_timezone`. This will make the line too long past the 120char limit, might need to do something like:
```
            sent_time_as_timezone = (sent_time.replace(tzinfo=dt.timezone.utc)
                                     .astimezone(self.protocol.timezone))
```

O365/message.py
```diff
@@ -755,7 +756,20 @@ def reply(self, to_all=True):
...
+            return '<b>Sent:</b> {:%A, %B %d, %Y %I:%M:%S %p}<br>'.format(sent_time_new)
```
Invincibear, line 767: Replace `sent_time_new` with `sent_time_as_timezone`

O365/message.py
```diff
@@ -755,7 +756,20 @@ def reply(self, to_all=True):
             return None
 
         message = response.json()
+        message_body_content_html  = message['body']['content']
 
+        pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
+
+        def change_sent_time(match):
```
Invincibear, line 763: Now that I understand the point of this PR, we should improve the naming of this function. We can name it `change_sent_time_timezone`, because technically we aren't changing the sent time itself, only the associated timezone

We should also add a quick comment about what this function does
```
"""Uses RegEx to extract the Sent time and attaches a timezone (self.protocol.timezone) to it"""
```

O365/message.py
```diff
@@ -755,7 +756,20 @@ def reply(self, to_all=True):
             return None
 
         message = response.json()
+        message_body_content_html  = message['body']['content']
```
Invincibear, line 759: You have an extra space before the `  = `

O365/message.py
```diff
@@ -755,7 +756,20 @@ def reply(self, to_all=True):
...
+
+        def change_sent_time(match):
+            sent_time_str = match.group(1)
+            sent_time = dt.datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
+            sent_time_new = sent_time.replace(tzinfo=dt.timezone.utc).astimezone(self.protocol.timezone)
+            return '<b>Sent:</b> {:%A, %B %d, %Y %I:%M:%S %p}<br>'.format(sent_time_new)
+
+        modified_data = pattern.sub(change_sent_time, message_body_content_html )
+
+        message['body']['content'] = modified_data
```
Invincibear, line 771: Per my previous comment, we can delete the above line and combine the two lines into the following:
```
# Inject original message body content's Sent time with a timezone
message['body']['content'] = pattern.sub(change_sent_time_timezone, message_body_content_html)
```

Invincibear (changes requested):

O365/message.py
Invincibear: Looks like you figured it out, let's just keep working at this, we can see the commit history of this PR here: https://github.com/O365/python-o365/pull/1043/commits

Invincibear:
If you can, you should rename the PR to something like `New feature: add timezone to message Sent time`

Elq-eng:
> If you can, you should rename the PR to something like `New feature: add timezone to message Sent time`

Ok, please check

O365/message.py
```diff
@@ -756,20 +756,20 @@ def reply(self, to_all=True):
             return None
 
         message = response.json()
-        message_body_content_html  = message['body']['content']
+        message_body_content_html = message['body']['content']
 
         pattern = re.compile(r'<b>Sent:</b> (.*?)<br>')
 
-        def change_sent_time(match):
+        """Uses RegEx to extract the Sent time and attaches a timezone (self.protocol.timezone) to it"""
```
Invincibear, line 763: This docstring is supposed to go inside the function, move it down one line

O365/message.py
```diff
@@ -756,20 +756,20 @@ def reply(self, to_all=True):
...
+        def change_sent_time_timezone(match):
             sent_time_str = match.group(1)
             sent_time = dt.datetime.strptime(sent_time_str, '%A, %B %d, %Y %I:%M:%S %p')
-            sent_time_new = sent_time.replace(tzinfo=dt.timezone.utc).astimezone(self.protocol.timezone)
-            return '<b>Sent:</b> {:%A, %B %d, %Y %I:%M:%S %p}<br>'.format(sent_time_new)
+            sent_time_as_timezone = (sent_time.replace(tzinfo=dt.timezone.utc)
+                            .astimezone(self.protocol.timezone))
```
Invincibear, line 768: so close, tab spacing here is off though, the `.` should line up just after the `(`. I'm surprised your IDE didn't catch this

Invincibear (changes requested):
hopefully last round of reviews

Invincibear:
> > If you can, you should rename the PR to something like `New feature: add timezone to message Sent time`
> 
> Ok, please check

I see that the commit name changed, thanks, you should also change the name of the PR though. If you go to this PR on GitHub at the top you should be able to rename it.

Also, after you make these final changes can you show a successful test of the method so we know it doesn't have any errors?

Elq-eng:
Well, please check out these pictures where the method is tested.


Before
after

Invincibear (approved):
LGTM

O365/message.py
alejcas: Exactly, no timezone hardcoded, no pytz usage

O365/message.py
alejcas: That's no longer True. Python 3.9 it's the oldest version possible

alejcas:
> LGTM

Means Looks Good To Me.
Common here on github

alejcas:
There are a few issues with this PR.

First, yo can not inject (nor search for) the "Sent" keyword. This could be "Enviado" if the user uses spanish...
So it will fail for other locales and also inject an incorrect keyword.

Other than that the problem with O365 it's that it was designed to always send to the server datetimes in UTC. That's why don't see the sent datetime like in your locale system settings.
I will try to fix this so your problem is fixed as well by default.

alejcas:
Thanks @Invincibear for the review process!

alejcas:
Thanks anyway for the effort!

Issue #1044, opened 2024-01-15
ZoneInfo object has no attribute Zone exception


geoffda:
windows_tz.py, line 638 is now throwing 'ZoneInfo object has no attribute Zone' exception when Calendar.new_event is called.

        iana_tz.zone if isinstance(iana_tz, tzinfo) else iana_tz)

Rolling back to o365 2.0.31 resolves the issue. I tried independently rolling back tzlocal to version 4.3.1, but apparently >=5.0 is required in the latest o365.

topcats:
hi there

yes they have changed the zoneinfo item

if you edit windows_tz.py and change the `iana_tz.zone` to `iana_tz.key` it works.
there are 2 places to change this

Invincibear:
> hi there
> 
> yes they have changed the zoneinfo item
> 
> if you edit windows_tz.py and change the `iana_tz.zone` to `iana_tz.key` it works. there are 2 places to change this

This worked for me, thanks!

MarcoMiano:
going back to tzlocal=4.3.1 works too...

alejcas:
Fixed some time ago

PR #1045, opened 2024-01-16
Fix broken references to iana_tz.zone in windows_tz.py

Invincibear:
This diff replaces broken `iana_tz.zone` references with working references to `iana_tz.key`, per the suggestion of @topcats in [this comment](https://github.com/O365/python-o365/issues/1044#issuecomment-1893460549). This resolved the 'ZoneInfo object has no attribute Zone' exception I was getting on my end.

This should resolve the following issues:
- https://github.com/O365/python-o365/issues/1011
- https://github.com/O365/python-o365/issues/1044

### DIFF
--- a/O365/utils/windows_tz.py
+++ b/O365/utils/windows_tz.py
@@ -635,9 +635,9 @@ def get_windows_tz(iana_tz):
     HOLY FUCKING SHIT!.
     """
     timezone = IANA_TO_WIN.get(
-        iana_tz.zone if isinstance(iana_tz, tzinfo) else iana_tz)
+        iana_tz.key if isinstance(iana_tz, tzinfo) else iana_tz)
     if timezone is None:
         raise ZoneInfoNotFoundError(
-            "Can't find Iana TimeZone " + iana_tz.zone)
+            "Can't find Iana TimeZone " + iana_tz.key)
 
     return timezone


RogerSelwyn:
This only works if you do not pass a timezone in on your account setup call:
`account = Account(credentials, token_backend=token_backend)`.

If you do the following it will fail:
`account = Account(credentials, token_backend=token_backend, timezone="UTC")` 
or `account = Account(credentials, token_backend=token_backend, timezone=timezone.utc)`

With the first call it will fail at line 100 of connection.py as noted in my first post on the linked issue, with the second call it will fail with missing 'key' value because the key is only available when you make the call to get_localzone() at line 102 of connection.py.

It will also fail in utils.py when you go to create a date to be sent to O365 in _build_time_time_zone which also references zone.

I was coming here to create a PR based on my last post on https://github.com/O365/python-o365/issues/1011. As noted there, it doesn't solve the issue as line 100 in connection.py, I'm not sure how that ever worked for the person who changed it in relation to the pytz removal.

Invincibear:
Let's put this on hold until I can devote a few minutes to working this out

alejcas:
This should be fixed in fd97b0fc8d2575c59e4e24585a9d3a7330d5e4e7

Please see #1051, a big change on date time timezone handling is coming

Issue #975, opened 2023-06-13
How to get mailboxSettings values?


sdelgadoc:
I have searched all over the code, and couldn't find a way to get the timeZone property on the mailboxSettings, which you can get with the following call:

`GET /me/mailboxSettings/timeZone`

You can find the documentation for the mailboxSettings here.
https://learn.microsoft.com/en-us/graph/api/resources/mailboxsettings?view=graph-rest-1.0

Any input on how to pull the information would be much appreciated.

janscas:
It's possible but not yet implemented

sdelgadoc:
Thank you, @janscas. Is another property implemented in a similar way that I could refer to in order to create a pull request (PR) for adding the timeZone property?

janscas:
Sorry there's already a MailboxSettings class in https://github.com/O365/python-o365/blob/master/O365/mailbox.py

See there what you want to add or change

PR #1050, opened 2024-01-27
Added time zone & working hours to MailboxSettings class

sdelgadoc:
#### Background
For some applications that rely on user information to provide functionality, it is helpful to retrieve user data, such as the user's time zone and working hours, directly from the Microsoft Graph API instead of asking the user for this information.

#### Challenge
The current implementation of MailboxSettings does not expose the user's  time zone & working hours even though the information is retrieved from the Microsoft Graph API by the code.

#### Changes
This PR exposes the timeZone and workingHours mailbox settings to the user as class variables in the MailboxSettings class.

#### Appendix
This functionality will help in the development of  [AdminGPT](https://github.com/sdelgadoc/AdminGPT) an AI-powered administrative assistant, implemented using OpenAI's Assistant framework to seamlessly integrate with your email and calendar.

Resolves O365/python-o365#975

### DIFF
--- a/O365/mailbox.py
+++ b/O365/mailbox.py
@@ -204,6 +204,8 @@ class MailboxSettings(ApiComponent):
         self.automaticrepliessettings = self.autoreply_constructor(
             parent=self, **{self._cloud_data_key: autorepliessettings}
         )
+        self.timezone = cloud_data.get("timeZone") 
+        self.workinghours = cloud_data.get("workingHours") 
 
     def __str__(self):
         """Representation of the MailboxSetting via the Graph api as a string."""


RogerSelwyn:
I would be worthwhile adding the ability to configure these via the api as well, it appears to be possible - https://learn.microsoft.com/en-us/graph/api/user-update-mailboxsettings?view=graph-rest-1.0&tabs=http

sdelgadoc:
> I would be worthwhile adding the ability to configure these via the api as well, it appears to be possible - https://learn.microsoft.com/en-us/graph/api/user-update-mailboxsettings?view=graph-rest-1.0&tabs=http

I think that's a really good idea! Would you like to collaborate on another pull request to implement that functionality? Because this is a very simple change, and it is important for an application that I'm developing, I'd like to keep this pull request to only this functionality.

RogerSelwyn:
It’s not something I need so I wouldn’t plan to create the PR.  I’m just considering that adding the update functionality looks pretty simple and would make this a complete implementation, rather than partial.

It’s not for me to approve, I’m just a contributor like you. @alejcas  will decide if he needs the extra.

alejcas:
Hi, merging this as it is.

Any improvement will be very welcome and merged.

Thanks for you efforts.

PR #1055, opened 2024-02-23
Add web_link property to Calendar

Invincibear:
Adds webLink property from Event spec: https://learn.microsoft.com/en-us/graph/api/resources/event?view=graph-rest-1.0


Invincibear:
I can confirm that this works, copy/pasted from my own modification to O365 used by my app until this PR is approved

alejcas:
thanks!

PR #1056, opened 2024-03-05
Expose task status

RogerSelwyn:
Currently only derived fields are exposed, e.g. is_complete. However status can have multiple values, so should be exposed directly.


alejcas:
This is included in 2.0.35 release

PR #1071, opened 2024-05-08
Update windows_tz.py

jackill88:
In some cases linux (on WSL2) uses "Europe/Kiev" instead of "Europe/Kyiv". It's the same time zone


alejcas:
Thanks!

PR #1083, opened 2024-06-05
added bucket_id to allowed update-attributes of Task

### LINKED ISSUE #1067: 'bucket_id' is missing in accepted update parameters
Could you add the 'bucket_id' in the accepted parameters for the update method? It is not possible to move tasks through buckets otherwise. Thanks!

https://github.com/O365/python-o365/blob/855dfc213a47ecea223a5d6a297d1711d11e4de7/O365/planner.py#L345

dekiesel:
Fixes #1067


alejcas:
Thanks!

PR #1105, opened 2024-10-02
fixing itemAt endpoint to new structure

ShadiNM:
Table Row operations are not aligned anymore with Microsoft Graph Docs 
https://learn.microsoft.com/en-us/graph/api/tablerowcollection-itemat?view=graph-rest-1.0&tabs=http

See also here, but this did not fully fix the issue
https://github.com/O365/python-o365/pull/849

could you please review and merge if this is fine?

### DIFF
--- a/O365/excel.py
+++ b/O365/excel.py
@@ -1334,7 +1334,8 @@ class Table(ApiComponent):
             return None
 
         url = self.build_url(self._endpoints.get('get_row_index'))
-        response = self.session.post(url, data={'index': index})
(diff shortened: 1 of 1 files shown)


ShadiNM:
@alejcas is everything ok with this change?

alejcas:
Thanks!

ShadiNM:
@alejcas thanks for merging! Could you let me know when this will be released to PyPI as version 2.0.37?
I would like to use the newest changes in my project

alejcas:
I've just released 2.0.37

PR #1110, opened 2024-11-06
User: add ability to access onPremisesSamAccountName

danpoltawski:
The User object has everything I need... except the ability to access the onPremisesSamAccountName to be able to integrate with on-prem applications.

I feel like I agree with [this comment](https://github.com/O365/python-o365/issues/298#issuecomment-528493118) of having a way of providing access to these sort of 'niche' attributes without needing to modify the object - perhaps just exposing a way to accessing the cloud data - but I can't see a way of achieving this with the current code, so here is a pull request to provide access to my new niche attribute :)


alejcas:
Thanks!

PR #1112, opened 2024-11-12
Add DjangoTokenBackend to Support Django Database for Token Storage

sdelgadoc:
**Background**
Python Django is a popular web development framework. It provides a flexible database interface that supports various types of databases. When developing Django applications, it is best practice to store non-volatile data in the Django database using its interface.

**Challenge**
Current TokenBackend implementations do not support the Django database interface directly, nor provide a compatible interface for use with Django.

**Changes**
This PR introduces the `DjangoTokenBackend` class, which extends `BaseTokenBackend` to support storing tokens in a Django database. It also updates `__init__.py` to make it easier for users to load `DjangoTokenBackend`.

**Appendix**
This functionality will assist in the development of [AdminGPT](https://github.com/sdelgadoc/AdminGPT), an AI-powered administrative assistant integrating seamlessly with email and calendar systems.


alejcas:
great!

sdelgadoc:
Thank you for the very quick response!

PR #1121, opened 2025-02-03
Added in README necessary step to set app to "Multitenant"

sdelgadoc:
I was creating a new app in Entra, and realized that one critical step was missing. This is a small change, but will likely stop folks from being able to setup the authentication.


alejcas:
Thanks!

PR #1124, opened 2025-02-11
Suggested changes to changlog

RogerSelwyn:
I've added some extra items which you can include or exclude as you like


alejcas:
These changes make perfect sense

PR #1137, opened 2025-02-28
Add email receipt notifications through webhooks

sdelgadoc:


sdelgadoc:
Thanks @alejcas.  I think that adding the three endpoints you listed is reasonable.  However, with the holidays coming up, I will likely not be able to start implementing until early to mid December.  So, depending on @kwilsonmg and other users' timeline, we can either merge this PR to give folks basic functionality, and I can add the additional endpoints later, or wait and merge all the functionality together in a few weeks.  Either works for me.

PR #1151, opened 2025-04-08
Docs tidy up

RogerSelwyn:
This should delete .doctrees and add it to .gitignore. Also a tidy up on account.rst because it didn't build.


RogerSelwyn:
If it doesn't completely delete .doctrees, might need to do a tidy up.

alejcas:
is it ready to merge?

RogerSelwyn:
Yes. But if it doesn't remove all .doctrees, you may need to do some tidy up.

alejcas:
ok, thanks!

alejcas:
These are the files included in the commit

image

alejcas:
I think only changes to docs\source should be committed

RogerSelwyn:
I think everything is sorted with that merge. It looks ok

PR #1152, opened 2025-04-15
Modify to handle threads using same account/connection object

RogerSelwyn:
I've made two modifications to this, one in connection.py and one in the examples. Both of these I have tested through, and without the changes I get 401 errors.

connection.py - I've changed the method of how the 401 error is ignored. This is because in my app, multiple threads are using the same account/connection object, so the existing flag was already set to True for any threads after the first that were doing a query. What I believe is required is that all queries get two chances in `oauth_request`, so the first call to `_internal_request` should 'ignore' a 401 and raise a TokenExpiredError, and the second call should raise the 401 error if repeated.

token_backends.py - The existing example did not loop (default max 3 times) while the second thread (when finding the file locked) waited for the token to be refreshed.

### DIFF
--- a/O365/connection.py
+++ b/O365/connection.py
@@ -581,14 +581,6 @@ class Connection:
             "https://login.microsoftonline.com/common/oauth2/nativeclient"
         )
 
-        # In the event of a response that returned 401 unauthorised this will flag between requests
-        # that this 401 can be a token expired error. MsGraph is returning 401 when the access token
-        # has expired. We can not distinguish between a real 401 or token expired 401. So in the event
-        # of a 401 http error we will first try to refresh the token, set this flag to True and then
-        # re-run the request. If the 401 goes away we will then set this flag to false. If it keeps the
-        # 401 then we will raise the error.
-        #: Indicates if the token has expired. |br| **Type:** bool
-        self._token_expired_flag: bool = False
 
     @property
     def auth_flow_type(self) -> str:
@@ -978,13 +970,14 @@ class Connection:
         self._previous_request_at = time.time()
 
     def _internal_request(
-        self, session_obj: Session, url: str, method: str, **kwargs
+        self, session_obj: Session, url: str, method: str, ignore401: bool, **kwargs
     ) -> Response:
         """Internal handling of requests. Handles Exceptions.
 
         :param session_obj: a requests Session instance.
         :param str url: url to send request to
         :param str method: type of request (get/put/post/patch/delete)
+        :param bool ignore401: indicates whether to ignore 401 error or not
         :param kwargs: extra params to send to the request api
         :return: Response of the request
         :rtype: requests.Response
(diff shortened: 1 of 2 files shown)


RogerSelwyn:
I've just tested this through in my test environment, and it works fine. I triggered an overlap by putting a delay prior to the refresh_token() in the LockableFSTB, this ensured following threads hit the file lock. These successfully looped until the refresh occurred (and lock removed) and new token loaded.

So I'm confident this is correct.

O365/connection.py
```diff
@@ -978,13 +970,14 @@ def _check_delay(self) -> None:
         self._previous_request_at = time.time()
 
     def _internal_request(
-        self, session_obj: Session, url: str, method: str, **kwargs
+        self, session_obj: Session, url: str, method: str, ignore401: bool, **kwargs
```
alejcas, line 973: Can you please define it a little bit different?

```python
def _internal_request(
        self, session_obj: Session, url: str, method: str, *, ignore401: bool = False, **kwargs
    ) -> Response:
```
This is just so if someone passes args to internal requests `ignore401` param don't receive an unintended param. Also we define a default value for it.

O365/connection.py
```diff
@@ -978,13 +970,14 @@ def _check_delay(self) -> None:
...
     ) -> Response:
         """Internal handling of requests. Handles Exceptions.
 
         :param session_obj: a requests Session instance.
         :param str url: url to send request to
         :param str method: type of request (get/put/post/patch/delete)
+        :param bool ignore401: indicates whether to ignore 401 error or not
```
alejcas, line 980: I think this docstring should indicate the real purpose of this param. This is whether or not try to interpret a 401 as a token expired error.

O365/connection.py
```diff
@@ -1103,7 +1095,7 @@ def naive_request(self, url: str, method: str, **kwargs) -> Response:
             # lazy creation of a naive session
             self.naive_session = self.get_naive_session()
 
-        return self._internal_request(self.naive_session, url, method, **kwargs)
+        return self._internal_request(self.naive_session, url, method, ignore401=False **kwargs)
```
alejcas, line 1098: I see not comma behind the param:

Like this:
`return self._internal_request(self.naive_session, url, method, ignore401=False, **kwargs)`

alejcas (commented):
I have added some comments. Let me know what you think, and change what best fits

O365/connection.py
RogerSelwyn: Added some extra test on latest commit.

O365/connection.py
alejcas: You didn't add the star before the ignore401, but anyway I've merged it

alejcas:
> Sorry... What does the star do? It is new to me. Do you want me to fix it?

The star ensures that the only way for the `ignore401` param  to receive a value is if the caller explicitly does mention it like this:

Without the star:

`result = self._internal_request(self.session, 'url...', True)`

With the star:

The previous call will raise an error.
The caller need to explicitly name it:

`result = self._internal_request(self.session, 'url...', ignore401=True)`

Issue #1086, opened 2024-06-11
CC email address not removed


dkatsios:
when an email reply is created using message.reply() if there is a CC recipient in the initial message, then even if reply.cc.clear() is called (or reply.cc.remove(the-email-address) when reply.send() is executed it will send the email also to the initial CC address.

alejcas:
try this:

```python
# ...
reply = message.reply(to_all=True)
reply.cc.clear()
reply.save()
reply.send()
```

RogerSelwyn:
PR submitted. No need to do save after the cc.clear. It should have been reply.save_message or reply.save_draft, but those are only needed if you want to save the draft for later.

PR #1153, opened 2025-04-16
Fix CC email address not remove

RogerSelwyn:
https://github.com/O365/python-o365/issues/1086


alejcas:
Yeah that was indeed an error I think

Issue #684, opened 2021-09-14
Selecting a field in a query breaks is_event_message and get_event()


antti-ngp:
Continuing on [683](https://github.com/O365/python-o365/issues/683)

I'm looping through messages in a mailbox and filtering results using a query to limit fields that I want:
```
q = mailbox.q().select('id')
for message in mailbox.get_messages(query=q):
    print(message.get_event())
    print(message.id)
```
I'm expecting to get the `id` of the message and all the data from the associated event.
Instead I get:
> None
> AA_MY_MESSAGE_ID_AAA=

I've found some [discussion](https://stackoverflow.com/questions/44954058/select-fields-on-microsoft-graph-list-of-messages) on what to include in the `select` to get the info whether a Message is a n EventMessage or not. This lead me into trying to include `microsoft.graph.eventMessage/meetingMessageType` in the `select` list which didn't work as there are multiple steps of handling done to the attributes before they are passed to the API. I tried modifying several of those steps to be able to pass the above attribute to the API with no luck. 

However, in the process, I discovered that commenting rows 630 to 634 in `/utils/utils.py`:
```
@fluent
    def select(self, *attributes):
        """ Adds the attribute to the $select parameter

        :param str attributes: the attributes tuple to select.
         If empty, the on_attribute previously set is added.
        :rtype: Query
        """
        if attributes:
            for attribute in attributes:
                attribute = self.protocol.convert_case(
                    attribute) if attribute and isinstance(attribute,
                                                           str) else None
                # if attribute:
                #     if '/' in attribute:
                #         # only parent attribute can be selected
                #         attribute = attribute.split('/')[0]
                #     self._selects.add(attribute)
        else:
            if self._attribute:
                self._selects.add(self._attribute)

        return self
```
gave me the expected results even when no additional fields were added to the `select`:
> Subject: My calendar event (on: 2021-09-27 from: 14:00:00 to: 16:00:00)
> AA_MY_MESSAGE_ID_AAA=

I don't understand how, but maybe `microsoft.graph.eventMessage/meetingMessageType` is somehow always implicitly included in the `select` and splitting on the character `/` somehow breaks it.

antti-ngp:
Having written that I now realise that what I've commented out basically equals to removing the whole `select` clause 😅 .

Did some further testing. Here's how to make it work:
1. Remove the hardcoded `camelcasing` from the `MSGraphProtocol` class on row 216 in `connection.py`
2. Pass `None` as the `casing_function` to the `protocol`
[...]

PR #1154, opened 2025-04-17
Add mapping for complex attributes

RogerSelwyn:
https://github.com/O365/python-o365/issues/684

In support of this issue, this adds the ability to convert a select attribute from a simple attribute to a complex one required by MS Graph.

### DIFF
--- a/O365/utils/utils.py
+++ b/O365/utils/utils.py
@@ -636,9 +636,6 @@ class Query:
         'flag': 'flag/flagStatus',
         'body': 'body/content'
     }
-    _select_mapping = {
-        "meetingMessageType": "microsoft.graph.eventMessage/meetingMessageType",
-    }
 
     def __init__(self, attribute=None, *, protocol):
         """ Build a query to apply OData filters
@@ -865,8 +862,11 @@ class Query:
         return None
 
     def _get_select_mapping(self, attribute):
-        mapping = self._select_mapping.get(attribute)
-        return mapping or attribute
+        if attribute in ["meetingMessageType"]:
+            return (
+                f"{self.protocol.keyword_data_store['event_message_type']}/{attribute}"
+            )
+        return attribute
 
     @fluent
     def new(self, attribute, operation=ChainOperator.AND):
--- a/O365/utils/utils.py
+++ b/O365/utils/utils.py
@@ -636,6 +636,9 @@ class Query:
         'flag': 'flag/flagStatus',
         'body': 'body/content'
     }
+    _select_mapping = {
+        "meetingMessageType": "microsoft.graph.eventMessage/meetingMessageType",
+    }
 
     def __init__(self, attribute=None, *, protocol):
         """ Build a query to apply OData filters
@@ -685,6 +688,7 @@ class Query:
                     if '/' in attribute:
                         # only parent attribute can be selected
                         attribute = attribute.split('/')[0]
+                    attribute = self._get_select_mapping(attribute)
                     self._selects.add(attribute)
         else:
             if self._attribute:
@@ -859,6 +863,10 @@ class Query:
                 attribute = self.protocol.convert_case(attribute)
             return attribute
         return None
+
+    def _get_select_mapping(self, attribute):
+        mapping = self._select_mapping.get(attribute)
+        return mapping or attribute
 
     @fluent
     def new(self, attribute, operation=ChainOperator.AND):


O365/utils/utils.py
```diff
@@ -636,6 +636,9 @@ class Query:
         'flag': 'flag/flagStatus',
         'body': 'body/content'
     }
+    _select_mapping = {
+        "meetingMessageType": "microsoft.graph.eventMessage/meetingMessageType",
```
alejcas, line 640: This is wrong as this info which uses a keyword referencing ms graph it's owned by the protocol.

This keyword is currently stored in the protocol.

Check this [line](https://github.com/O365/python-o365/blob/master/O365/utils/query.py#L734) for example. 

So instead it should do something like this:

```python
attribute = f"{self.protocol.keyword_data_store['event_message_type']}/meetingMessageType"
```

alejcas (commented):
First I have write a new Query class that I want to replace the old one.
You can check it [here](https://github.com/O365/python-o365/blob/master/O365/utils/query.py).

Also I've added a comment on a line that needs to change.

O365/utils/utils.py
RogerSelwyn: OK, this depends on new query class. I think I should wait until you have that in place, since you will be changing other calls to it. I can redo the PR when it is ready.

RogerSelwyn: Or I can make it inline with the old version of the class, which I can see handles it in 'expand'.

alejcas: Whatever you prefer. Anyway you can add this also to the new Query class

RogerSelwyn: > Whatever you prefer. Anyway you can add this also to the new Query class

I've left it in Utils, let me know if it should be move to Query. Where you linked to in Query was for an Expand, whilst this is a select.

alejcas: This should be added as well for the new query.

RogerSelwyn: I'm not sure how you mean to include it there, do you mean in here:

```python
    def select(self, *attributes: str) -> SelectFilter:
        """
        Returns a 'select' query param
        This is useful to return a limited set of attributes from a resource or return attributes that are not
         returned by default by the resource.

        :param attributes: a tuple of attribute names to select
        :return: a SelectFilter instance that can render the OData select operation
        """
        select = SelectFilter()
        for attribute in attributes:
            select.append(self.protocol.convert_case(attribute))
        return select
```

Not sure how to test it, but I can add the same logic. Shall I open a new PR?

alejcas:
Also, I will be thankful if you can try the new Query.
It works like this:

```python
# here you have you account set up:
from O365.utils import ExperimentalQuery

builder = ExperimentalQuery(protocol=account.protocol)

query = builder.equals("name", "James") | builder.equals("address", "james@example.com")
query = query & builder.orderby("name", "address")

# then use query as where the old query was used in the same way
```

For example to add select just mix whatever part with `builder.select` with the & operator

O365/utils/utils.py
alejcas: yeah it should be there. But I think you can also expand on this attribute?
Maybe this should work both for select and expand.

Also note that I think it's better if the check is done ignoring the casing of the provided attribute like:

`if attribute.lower() == "meetingmessagetype"`

alejcas:
> Also, I will be thankful if you can try the new Query. It works like this:

The reasons for a new query builder are:

- Allows for more deep operations like using iterator like any or all over complex attributes or deep collections
- It's way more simple and error prone.
- It allows for the use of `&` and `|` operators (but also implements `builder.chain_and/chain_or")
- It allows to group filters explicitly in the way you like with `builder.group`
- It properly manages the negate condition being explicit about it's use with `builder.negate`

O365/utils/utils.py
RogerSelwyn: > But I think you can also expand on this attribute?

No - this error

`"Parsing OData Select and Expand failed: Property 'meetingMessageType' on type 'microsoft.graph.eventMessage' is not a navigation property or complex property. Only navigation properties can be expanded."`

alejcas: > > But I think you can also expand on this attribute?
> 
> No - this error
> 
> `"Parsing OData Select and Expand failed: Property 'meetingMessageType' on type 'microsoft.graph.eventMessage' is not a navigation property or complex property. Only navigation properties can be expanded."`

oh. Ok!!

PR #1163, opened 2025-04-30
Remove Office 365 Protocol and documentation

RogerSelwyn:
Office 365 api is no linger available from Microsoft, so this PR is to remove it. If you disagree, then feel free to reject.

I've also change mention of Office 365, (where it does not relate to the Office 365 API), to Microsoft 365 in line with current MS branding.

### DIFF
--- a/O365/__init__.py
+++ b/O365/__init__.py
@@ -1,5 +1,5 @@
 """
-A simple python library to interact with Microsoft Graph and Office 365 API
+A simple python library to interact with Microsoft Graph
 """
 import warnings
 import sys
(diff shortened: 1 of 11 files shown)


alejcas:
ok. It's true Office 365 must be removed as it's deprecated.
But I want to also reflect that O365 can be implemented with other apis like microsoft sharepoint rest api or ms dynamics rest api.

Can you not change the tittle of the library or at least leave something like:

"A simple python library to interact with Microsoft Graph and other MS api"

RogerSelwyn:
> ok. It's true Office 365 must be removed as it's deprecated. But I want to also reflect that O365 can be implemented with other apis like microsoft sharepoint rest api or ms dynamics rest api.
> 
> Can you not change the tittle of the library or at least leave something like:
> 
> "A simple python library to interact with Microsoft Graph and other MS api"

The other api I can see in the connection.py is MS Business Central 365, which I don't really know anything about. I assume this is Dynamics. I'm happy to go through and mention it into the docs in places. I'm not sure which objects it supports access to and there isn't anything in the docs that mentions it to go on so it would be very light.

You should be able to commit to this PR as well to improve it.

RogerSelwyn:
Or feel free to reject it, if it isn't setting the right impression, I'm very relaxed about it :-)

PR #1173, opened 2025-07-07
Added docs for query builder

RogerSelwyn:
Added docs for query builder based on old query helper docs. Two comments to make:

1. Should this still be called ExperimentalQuery if it is now the recommended way of creating queries?
2. Previously you could instantiate a query with a simple `mailbox.new_query`. Now you must:

```python
from O365.utils import ExperimentalQuery
builder = ExperimentalQuery(protocol=account.protocol) # Which assume you have account available when needed
query = builder.xxxxxx
```
  Would it be feasible to add the new builder method to ApiComponent?

### DIFF
--- a/O365/excel.py
+++ b/O365/excel.py
@@ -1954,7 +1954,7 @@ class WorkSheet(ApiComponent):
         direct endpoint API for tableless row updates.
         :param str|Range address: the address to resolve to a range which can be used for updating cells.
         :param list[list[str]] rows: list of rows to push to this range. If updating a single cell, pass a list
-            containing a single row (list) containing a single cell worth of data.
+        containing a single row (list) containing a single cell worth of data.
         """
         if isinstance(address, str):
             address = self.get_range(address)
--- a/docs/source/usage/sharepoint.rst
+++ b/docs/source/usage/sharepoint.rst
@@ -11,8 +11,8 @@ Sites.Read.All             sharepoint                               To only read
 Sites.ReadWrite.All        sharepoint_dl                            To read and save sites, lists and items
 =========================  =======================================  =======================================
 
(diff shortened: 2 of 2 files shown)


alejcas:
There is not an easy way of introducing a breaking change… It should be experimental and advised for some time before changing the whole api. Otherwise I guess we’re making it difficult for users.

I think we can remove the old query and change the new one so it’s the default soon anyway

alejcas:
> ilder like yo

I'll change the default Query to the new one for the moment but keep the old one as well

RogerSelwyn:
@alejcas latest commit on here should build pages correctly when merged. I can move to separate PR if required.

RogerSelwyn:
I noted these errors in my test build. I can fix them in this or a separate PR if needed.

```
/home/runner/work/python-o365/python-o365/O365/excel.py:docstring of O365.excel.WorkSheet.update_cells:5: ERROR: Unexpected indentation. [docutils]
/home/runner/work/python-o365/python-o365/docs/source/usage/sharepoint.rst:14: WARNING: Inline emphasis start-string without end-string. [docutils]
```

alejcas:
> I noted these errors in my test build. I can fix them in this or a separate PR if needed.
> 
> ```
> /home/runner/work/python-o365/python-o365/O365/excel.py:docstring of O365.excel.WorkSheet.update_cells:5: ERROR: Unexpected indentation. [docutils]
> /home/runner/work/python-o365/python-o365/docs/source/usage/sharepoint.rst:14: WARNING: Inline emphasis start-string without end-string. [docutils]
> ```

Wow, I don't understand this. Seems like the indentation is correct (or as correct as in other methods in the same file)
Do you know where the problem is?

RogerSelwyn:
You should be good to go now

PR #1208, opened 2026-01-28
allow delay delivery option of email

yeyeric:
that feature was missing, tested and working perfectly. Either you provde an int of seconds to postpone delivery of emails, or absolute datetime properly formated in string as expected by microsoft


alejcas:
Thanks!

